PR #41, opened 2020-04-01
Invitation workflow for OCM

LovisaLugnegard:
This PR contains an invitation workflow for OCM to discover users in remote system while preserving privacy.


LovisaLugnegard:
@ishank011 and @labkode Do you see a reason for using both "invite" and "invitation" in the API description? If we use both I think we should be clear that they either mean the same (no need to use both in that case), or what the difference is.

spec.yaml
```diff
@@ -355,6 +355,47 @@ paths:
...
+        200:
+          description: Invitation forwarded.
+        400:
+          description: |
+            Bad request due to invalid parameters, e.g. provider is not valid.
+          schema:
+            $ref: "#/definitions/400"            
+  /invites/accept:
+    post:
+      summary: Accept an invititation to start sharing.
```
ishank011: typo

spec.yaml
```diff
@@ -505,4 +546,34 @@ definitions:
...
+      - properties:
+          token:
+            type: string
+            description: Token generated by the service sending the invitation.
+            example: xyz
+          senderProvider: 
+            type: string 
+            format: url
+            description: URL of the OCM sender service.
+            example: https://sender.org/ocm/
```
ishank011: Remove /ocm. We're not passing the endpoint, just the provider.

ishank011 (changes requested):
Hi Lovisa. Looks good - just a couple of comments.

spec.yaml
```diff
@@ -505,4 +546,34 @@ definitions:
...
+      - properties:
+          userID:
+            type: string
+            description: Unique ID to identify the user at the remote provider accepting the invite.
+            example: 9303
+          recipientProvider:
+            type: string 
+            format: url
+            description: URL of the receiver OCM service.
+            example: https://receiver.org/ocm
```
ishank011: Same. Remove /ocm

spec.yaml
```diff
@@ -355,6 +355,47 @@ paths:
               type: string
           schema:
             $ref: "#/definitions/Error"
+  /invites/forward:
+    post:
+      summary: Forward an invitation to start sharing
+      description: Forwards an invitation received by an out-of-band channel, email for example.
```
pferreir: Maybe "**through** an out-of-band channel", or "via" would be easier to understand?

LovisaLugnegard: That's a nice improvement, I'll update.

spec.yaml
```diff
@@ -355,6 +355,47 @@ paths:
               type: string
           schema:
             $ref: "#/definitions/Error"
+  /invites/forward:
```
michielbdejong, line 358: Who forwards the invitation when this end-point is called, and from where to where?

michielbdejong: IIUC from https://www.youtube.com/watch?v=TnlUOG3lOu4, the flow would be as follows, if Alice at server A invites Bob at server B:
* Alice instructs A to send an invite to "Bob@B"
  * A changes state, "invite sent from Alice to Bob@B"
  * A calls B/invites/forward
  * B changes state, "invite received from Alice@A to Bob"
* B notifies Bob (this is the "forwarding", right?)
* Bob instructs B to accept the invitation
  * B changes state, "invite from Alice@A to Bob accepted"
  * B calls A/invites/accept
  * A changes state, "invite from Alice to Bob@B accepted"

If this is correct, then I think the summary and description should reflect that it is B who forwards the invite from A (or effectively from Alice@A) to Bob?

michielbdejong: The summary "Forward an invitation to start sharing" suggests that the agent doing the forwarding (B) is also the agent who starts sharing, but I think it's more accurate to say Alice@A starts sharing.

Also, since B forwards from the inner network to the last mile end-point, "deliver" might have been a better verb than "forward".

The delivery of the notification is not the main functionality exposed by this API endpoint; rather, if and how B notifies Bob is not even the concern of A or of Alice@A. So given that there is also a state change in B, maybe "create" (as in "create an entry in your list of received invites") would have been even better than "forward" or "deliver".

The description "Forwards an invitation received through an out-of-band channel, email for example." describes how the invitation was received, but I think the intention was to describe how the invitation gets forwarded.
So maybe "Forwards a received invitation through an out-of-band channel, email for example." is clearer?

michielbdejong: And for the summary I would suggest "Notify the recipient of a new invite"

michielbdejong: Ah wait, I ran through https://reva.link/docs/tutorials/share-tutorial/#4-2-accept-the-token again and this actually looks like accepting the invite is not a server-to-server interaction, but a client-to-server one?
Should the OCM API not only define those, or leave that to each implementer?
In the end it doesn't matter to Alice or to A how Bob and B communicate with each other, right?

michielbdejong:
What is the difference between [ForwardInvite and AcceptInvite](https://github.com/cs3org/cs3apis/blob/414d8eb/cs3/ocm/invite/v1beta1/invite_api.proto#L56-L59)?

michielbdejong:
I think we should only add `/invites/accept` to the OCM API, and there is no need to define an `/invites/forward` endpoint.

Looking at the reva code [there is a `ForwardInvite` function](https://github.com/cs3org/reva/blob/v1.14.0/pkg/ocm/invite/manager/json/json.go#L219) which does a http call to 
[`acceptInviteEndpoint` "/invites/accept"](https://github.com/cs3org/reva/blob/v1.14.0/pkg/ocm/invite/manager/json/json.go#L47).

But [the `AcceptInvite` function](https://github.com/cs3org/reva/blob/v1.14.0/pkg/ocm/invite/manager/json/json.go#L246-L274) does not do any http calls (it's the function that *gets* called when a http request comes in over the server's OCM API.

michielbdejong:
I'll see if I can create a PR to supersede this one (which is now 18 months old and hasn't been getting much movement recently).

michielbdejong:
I think we should also define this end point as `application/json` (like the other ones), not as `application/x-www-form-urlencoded` which is currently the only format supported by reva.

michielbdejong:
I see revad additionally has a http endpoint where an authenticated user can cause an [invite to be generated and emailed out](https://github.com/cs3org/reva/blob/v1.14.0/internal/http/services/ocmd/invites.go#L73-L126).
There also is a [`forwardInvite` handler](https://github.com/cs3org/reva/blob/v1.14.0/internal/http/services/ocmd/invites.go#L128-L181) which takes an invite `token` and an origin `providerDomain` and converts those into a grpc request for `ForwardInvite`. It is unclear to me why a reva user would use this http method instead of just using the equivalent grpc method directly?

glpatcern:
Superseded by #54, please let's resume the discussion there.

PR #54, opened 2022-06-21
Invitation workflow for OCM, take 2

glpatcern:
This PR supersedes https://github.com/cs3org/OCM-API/pull/41.

I have rebased the new endpoints on top of the current `develop` branch. We should have a discussion with @michielbdejong as a main contributor in this repo also to address the yet outstanding comments from the previous PR, so this is still WIP though Reva currently includes a full implementation of these specs.


michielbdejong:
@glpatcern great, let's have a video call to walk through the details!

glpatcern (commented):
(Self-reviewing for further brainstorming)

spec.yaml
```diff
@@ -259,7 +295,9 @@ definitions:
       resourceType:
         type: string
         description: |
-          Resource type (file, calendar, contact,...)
+          Resource type (file, webapp, calendar, contact, ...). A `webapp` share
+          is an application that can be opened on the remote end, typically
+          to enable collaborative workflows.
```
glpatcern: This is a (non-binding) idea to further diversify the type of resource one is sharing, including the application engine to use to edit some files.

However, the action of opening a file in a web app would come _after_ the file was already shared with an `/ocm/share` call (and `resourceType=file`), so I'm not entirely sure an additional `/ocm/share` call with `resourceType=webapp` is the right semantic here.

That said, if the two EFSS both use Reva as the reference implementation of OCM, then the `open-in-app` workflow implemented in Reva ought to actually work out of the box, provided that an option (e.g. in the context menu) is given to the user in the web frontend to execute a "remote open" as opposed to a "local open". That hook would be served by the local Reva, which will forward it to the remote Reva as it happens on an OCM share: and the remote Reva will ultimately perform the "remote open", returning the app URL.

spec.yaml
```diff
@@ -140,15 +137,54 @@ paths:
...
-
                 Internet Message Format [RFC5322] (e.g. `Wed, 21 Oct 2015
                 07:28:00 GMT`) or the number of seconds 
-
                 (e.g. 3000 if you the service is expected to be available again
                 within 50 minutes).
               type: string
           schema:
             $ref: '#/definitions/Error'
+  /invites/forward:
```
glpatcern: For the record, there was also the idea to rename `/forward` as well, but I'm not sure something like `/deliver` would be much better.

On the other hand, I support renaming `/accept` to `/accepted` as this is really a message sent back to inform that the invitation was accepted.

spec.yaml
glpatcern: Resuming the discussion and targeting a resolution at the upcoming workshop @ CERN:

Even assuming that all EFSSs use Reva as the reference implementation of OCM (which is the target model in ScienceMesh), it would still be good to expose an `/ocm/open` endpoint, which Reva would simply forward to `/app/open`. In this context, a `webapp` resource type is indeed redundant, but on the OCM share part some `edit in app` kind of permission may be envisaged to enable such workflow, where the "normal" r/w permission would simply allow to download/upload data to the share.

smesterheide:
I can't speak to the issue at hand, but I am happy to see progress in advancing the spec to a new version. Is there a process to follow?

glpatcern:
> I can't speak to the issue at hand, but I am happy to see progress in advancing the spec to a new version. Is there a process to follow?

Hi @smesterheide, if you have some specific suggestion or RFE you are welcome to file an issue or a PR in GitHub. More in general the [CS3 community](https://www.cs3community.org/) regularly meets and discusses OCM issues among other topics, and I expect we'll tag a new version of the API at next event.

glpatcern:
@michielbdejong @labkode I propose to get this fully sorted out and merged in these days. At least it would reflect what is already (partially) implemented in Reva and in the OC10/NC integrations, as far as I understand.

spec.yaml
glpatcern: Discussing with @michielbdejong, the `forward` action is actually called `accept` in the OC/NC apps, and there's no explicit OCM `accept` action in those apps.

glpatcern:
@michielbdejong and @labkode I've tried to spell out what we brainstormed yesterday (and thanks for the nice discussion!). Please have a look so we can move forward with the necessary changes in Reva.

spec.yaml
```diff
@@ -326,3 +341,57 @@ definitions:
         example:
           message: "Recipient accepted the share"
           sharedSecret: "hfiuhworzwnur98d3wjiwhr"
+  AcceptedInvite:
```
michielbdejong, line 344: Where is this used?

spec.yaml
```diff
@@ -139,16 +141,37 @@ paths:
...
+                available again within 50 minutes).
               type: string
           schema:
             $ref: '#/definitions/Error'
+  /invite-accepted:
+    post:
+      summary: Inform the sender that the invitation was accepted to start sharing.
+      description: Inform about an accepted invitation so the user on the sender provider's side can initiate the OCM share creation.
+      parameters:
+        - name: invite
```
michielbdejong, line 156: IIUC you're proposing two deviations from the current implementation as it exists in Reva:
1. change the endpoint from e.g. `https://sender.com/iop/ocm/invite/accept` to e.g. `https://sender.com/iop/ocm/invite-accepted` -> as agreed, that makes it clearer
2. change the JSON from:
```
{
  "token": "10431cd1-11df-49ce-af87-9dfc",
  "recipientProvider": "surf-iop-rd-app-acc.data.surfsara.nl",
  "userID": "marie",
  "name": "Marie Curie",
  "email":  "marie@surfsara.nl"
}
```
to (IIUC):
```
{
  "invite" : {
    "token": "10431cd1-11df-49ce-af87-9dfc",
    "recipientProvider": "surf-iop-rd-app-acc.data.surfsara.nl",
    "userID": "marie",
    "name": "Marie Curie",
    "email":  "marie@surfsara.nl"
  }
}
```
Can this second deviation be avoided? Does it serve a purpose?

michielbdejong (changes requested):

spec.yaml
glpatcern: Interesting, this is a left over from the previous PR. I'll drop it.

spec.yaml
glpatcern: This was also not intended and comes from the original PR. Will fix it.

Edit: after deeper analysis it's good as is - thanks Michiel for double checking!

michielbdejong: As discussed on Gitter, see https://swagger.io/docs/specification/2-0/describing-request-body/ - 
> The payload name. It is required but ignored (it is used for documentation purposes only).

So the existing current format for `/shares` is (confusing but) correct, and so is yours here!

michielbdejong: Sorry for the confusion, you don't need to change anything.

michielbdejong (approved):
I see you removed the spurious section and the 'name' in OpenAPI 2.0 does *not* add a level of nesting, so all good!

Issue #49, opened 2021-06-01
NewShare field 'permission' is required but not defined


michielbdejong:
See https://github.com/cs3org/OCM-API/blob/develop/spec.yaml#L189-L296
There is  mention of `protocol.options.permissions` (plural) in https://github.com/cs3org/OCM-API/blob/develop/spec.yaml#L296 but no mention of what the root field `permission` should look like.

glpatcern:
Closed by #57

Issue #51, opened 2021-08-06
Endpoint discovery through https://example.com/ocm-provider/


michielbdejong:
In practice, if servers A and B participate in the OpenCloudMesh, and Alice@A wants to share a resource with Bob@B, then the first thing server A will do is to do a GET to `https://B/ocm-provider/`, for end-point discovery.

I don't think this practice is currently documented in this spec repo, but without it, OCM doesn't work.

There is another option - we could define endpoint discovery to be out-of-scope, meaning OpenCloudMesh will only work between previously federated servers. This is for sure how it will be used in ScienceMesh, so there it's fine, but in general, I think it would be a pity to lose the open-ended host discovery.

michielbdejong:
Unlike Nextcloud and ownCloud, reva uses a providers file (whitelist including config for each provider).

glpatcern:
Following the CS3 event in Barcelona, there are two points that came up on this, prior to merging #59:

1. Do we agree to expose on the `resourceTypes.protocols` list the top-level URL endpoints for each protocol? Which implies that the share action would provide URLs relative to those endpoints? The advantage for this scheme - according to Nextcloud - is to allow for an EFSS instance to move e.g. to a different domain without loosing all established shares. Of course, this would also imply that the discovery endpoint MUST be called by the EFSSs every now and then to refresh such top-level URLs. Note that if we go this direction, also ScienceMesh would depend on the discovery endpoint - which I think it's fine.

2. We should make it clear that implementations may provide any `resourceType`, `shareType`, and `protocol` as optional capabilities, whilst stating that the minimum requirement for an OCM-compliant implementation is to have a given minimum set of capabilities. IMHO a reasonable minimum would be `resourceType` = `file`, `shareType` = `user`, `protocol` = `webdav`. So far this is not mentioned anywhere, as apparently there were reasons to not "impose" `webdav` as minimum common protocol. Time to change?

smesterheide:
> 1. Do we agree to expose on the `resourceTypes.protocols` list the top-level URL endpoints for each protocol? Which implies that the share action would provide URLs relative to those endpoints? The advantage for this scheme - according to Nextcloud - is to allow for an EFSS instance to move e.g. to a different domain without loosing all established shares. Of course, this would also imply that the discovery endpoint MUST be called by the EFSSs every now and then to refresh such top-level URLs. Note that if we go this direction, also ScienceMesh would depend on the discovery endpoint - which I think it's fine.

[...]

Issue #23, opened 2016-12-12
Security proposal: trusted services and shared secrets / request signing


joostfarla:
In the file sharing context, security is a serious concern. Therefore, it's a best practice to keep your
[attack surface](https://en.wikipedia.org/wiki/Attack_surface) as small as possible. There are various measures to take when it comes to minimizing risk as much as possible.

## TLS/SSL

The TLS/SSL protocol (serving traffic over HTTPS) guarantees, when implemented correctly, an end-to-end secure connection. Support for SSL is an absolute MUST for any decent API nowadays. SSL helps encrypting any private data when transferring it over the wire. However, as we will see in the "Request signing" chapter, SSL only is not always enough.

## Trust relationships

To be able to initiate resource sharing between different users on different instances, some API endpoints have to be exposed. These endpoints, especially for open-source projects, are an easy target for malicious services and should be secure and only be accessible by services you trust. Therefore, the concept of "trusted services" could be implemented. When you configure a trusted service, you are *whitelisting* this service to access the endpoints used for the negotiation and revocation of resource shares.

A trusted service means you trust the owner of the service and all code in the service application
to behave itself at all times. You must also be confident that the trusted application will maintain
the security of its shared secret.

## Request signing

A common approach for managing trust relationships between services is by providing a
[shared secret](https://en.wikipedia.org/wiki/Shared_secret), which is generally a long sequence of randomly generated characters, known only to the parties involved. This shared secret could be generated by the *trusting* service admin and should be provided to the *trusted* party (to be). The  shared secret should be provided to the latter party once and not change over time (of course you could decide to *rotate* all shared secrets once in a while, but we leave this out of scope for now).

Even when using SSL, there is still the small chance of a [man-in-the-middle attack](https://en.wikipedia.org/wiki/Man-in-the-middle_attack). Recent OpenSSL exploits (e.g. Heartbleed) have also demonstrated that the SSL protocol could not be  considered 100% secure. Especially because different sharing services of different vendors could  be hosted in various different environments with different security risks/policies and management levels, it's not a wise decision to fully trust on SSL encryption. Therefore, a proper choice would be to use the shared secret for *signing* requests.

[...]

PR #57, opened 2023-01-16
Improved properties of the NewShare endpoint + some more cleanup

### LINKED ISSUE #56: Cannot specify options per protocol in create share endpoint
As https://github.com/cs3org/OCM-API/pull/54 introduced in the `/shares` endpoint the possibility of specifying multiple protocols, there is still no way of specifying the options per protocol, or at least is not clear.

This should be a map protocol -> options, for example
```
{
  "webdav": {
     "sharedSecret": "hfiuhworzwnur98d3wjiwhr",
    "permissions": "{https://open-cloud-mesh.org/ns}share-permissions",
    "uriTemplate": "[https://open-cloud-mesh.org/s/{path-to-shared-resources}](https://open-cloud-mesh.org/s/%7Bpath-to-shared-resources%7D)"
  },
  "other": {...}
}
```

gmgigi96:
Closes https://github.com/cs3org/OCM-API/issues/56

 Co-authored with @glpatcern

Includes:
* Properties of the `protocols` parameter in `NewShare` fully spelled out [this is a breaking change].
* Renamed `providerId` as `shareId`, both in `NewShare` and in `/notifications`, as this is how it is meant to be [this is a breaking change, though semantic is unchanged].
 
The corresponding changes in the CS3 APIs are in https://github.com/cs3org/cs3apis/pull/199

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -283,14 +283,14 @@ definitions:
           Other custom protocols might be added in the future.
         example:
           webdav:
-            sharedSecret: "hfiuhworzwnur98d3wjiwhr"
+            sharedSecret: "secret"
             permissions: ["read", "write", "share"]
             uri: "https://open-cloud-mesh.org/remote.php/webdav/path/to/resource"
           webapp:
-            sharedSecret: "hfiuhworzwnur98d3wjiwhr"
-            uriTemplate: "https://open-cloud-mesh.org/s/rglsertgsrt/{relative-path-to-shared-resource}"
+            sharedSecret: "secret"
+            uriTemplate: "https://open-cloud-mesh.org/s/{share-id}/{relative-path-to-shared-resource}"
           datatx:
-            sharedSecret: "hfiuhworzwnur98d3wjiwhr"
+            sharedSecret: "secret"
             srcUri: "https://open-cloud-mesh.org/remote.php/webdav/path/to/resource"
             size: 1000000000000
   NewNotification:
--- a/spec.yaml
+++ b/spec.yaml
@@ -48,7 +48,16 @@ paths:
       responses:
         "201":
           description: >
-            Consumer successfully received the share.
+            Consumer successfully received the share. The response might contain
+            the display name of the recipient of the share for general
+            user experience improvement.
+          schema:
+            type: object
+            properties:
+              recipientDisplayName:
+                type: string
+                description: display name of the recipient
+                example: John Doe
         "400":
           description: >
             Bad request due to invalid parameters, e.g. when `shareWith` is not
@@ -202,6 +211,7 @@ definitions:
       - shareWith
       - name
       - resourceId
+      - providerId
       - owner
       - sender
       - protocols
@@ -232,6 +242,12 @@ definitions:
         description: >
           Identifier to identify the resource at the provider side. This is
           unique per provider.
+        example: 7c084226-d9a1-11e6-bf26-cec0c932ce01
+      providerId:
+        type: string
+        description: >
+          Identifier to identify the resource at the provider side. This is
+          unique per provider (deprecated).
         example: 7c084226-d9a1-11e6-bf26-cec0c932ce01
       owner:
         description: |
--- a/spec.yaml
+++ b/spec.yaml
@@ -151,7 +151,11 @@ paths:
   /invite-accepted:
     post:
       summary: Inform the sender that the invitation was accepted to start sharing.
-      description: Inform about an accepted invitation so the user on the sender provider's side can initiate the OCM share creation.
+      description: >
+        Inform about an accepted invitation so the user on the sender provider's side can initiate the OCM share creation.
+        To protect the identity of the parties, for shares created following an OCM invitation,
+        the user id SHALL be hashed, and recipients implementing the OCM invitation workflow
+        MAY refuse to process shares coming from unknown parties.
       parameters:
         - name: invite
           in: body
@@ -226,7 +230,7 @@ definitions:
           to share the resource with. This is known in advance.
           Please note that the consumer service endpoint is known in advance
           as well, so this is no part of the request body.
-        example: peter.szegedi@geant.org
+        example: 51dc30ddc473d43a6011e9ebba6ca770@geant.org
       name:
         type: string
         description: Name of the resource (file or folder).
@@ -253,7 +257,7 @@ definitions:
         description: |
           Provider specific identifier of the user who owns the resource.
         type: string
-        example: dimitri@apiwise.nl
+        example: 6358b71804dfa8ab069cf05ed1b0ed2a@apiwise.nl
       sender:
         description: >
           Provider specific identifier of the user that wants to share the
@@ -261,7 +265,7 @@ definitions:
           identified on a higher level, so the former `remote` property
           is not part of the request body.
         type: string
-        example: john@apiwise.nl
+        example: 527bd5b5d689e2c32ae974c6229ff785@apiwise.nl
       ownerDisplayName:
         type: string
         description: |
@@ -394,7 +398,7 @@ definitions:
           userID:
             type: string
             description: Unique ID to identify the user at the remote provider accepting the invite.
-            example: 9303
+            example: 51dc30ddc473d43a6011e9ebba6ca770
           email:
             type: string
             description: Email ID of the user accepting the invite.
--- a/spec.yaml
+++ b/spec.yaml
@@ -244,7 +244,8 @@ definitions:
         type: string
         description: >
           Identifier to identify the shared resource at the provider side. This is
-          unique per provider and per share.
+          unique per provider such that if the same resource is shared twice,
+          this shareId will not be repeated.
         example: 7c084226-d9a1-11e6-bf26-cec0c932ce01
       owner:
         description: |
@@ -366,7 +367,7 @@ definitions:
         example: file
       shareId:
         type: string
-        description: ID of the shared resource on the provider side
+        description: ID of the shared resource on the provider side, formerly known as providerId
         example: 7c084226-d9a1-11e6-bf26-cec0c932ce01
       notification:
         type: object
--- a/spec.yaml
+++ b/spec.yaml
@@ -201,8 +201,9 @@ definitions:
     required:
       - shareWith
       - name
-      - providerId
+      - resourceId
       - owner
+      - sender
       - protocols
       - permission
       - shareType
@@ -226,9 +227,9 @@ definitions:
         example: >-
           This is the Open API Specification file (in YAML format) of the Open
           Cloud Mesh API.
-      providerId:
+      resourceId:
         type: string
-        description: >-
+        description: >
           Identifier to identify the resource at the provider side. This is
           unique per provider.
         example: 7c084226-d9a1-11e6-bf26-cec0c932ce01
--- a/spec.yaml
+++ b/spec.yaml
@@ -48,16 +48,7 @@ paths:
       responses:
         '201':
           description: >
-            Consumer successfully received the share. The response might contain
-            the display name of the recipient of the share for general
-            user experience improvement
-          schema:
-            type: object
-            properties:
-              recipientDisplayName:
-                type: string
-                description: display name of the recipient
-                example: John Doe
+            Consumer successfully received the share.
         '400':
           description: >
             Bad request due to invalid parameters, e.g. when `shareWith` is not
@@ -267,7 +258,7 @@ definitions:
       shareType:
         type: string
         description: |
-          Share type (user or group share)
+          Recipient share type (user or group)
         example: user
       resourceType:
         type: string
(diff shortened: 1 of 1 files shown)


glpatcern:
@michielbdejong we're now trying to implement this "for real", what do you think about this?

The example can surely be made with no repetitions

glpatcern:
For reference, OCM permissions are handled as follows in nextcloud:

https://github.com/nextcloud/server/blob/1021c894542a778948bfddb9722e10dddf91f4b8/apps/federatedfilesharing/lib/OCM/CloudFederationProviderFiles.php#L694-L709

glpatcern:
@michielbdejong and @labkode we should be good to go with those changes, could you please review?

This also closes https://github.com/cs3org/OCM-API/issues/49, by using the explicit list of strings adopted by Nextcloud (cf. my previous comment).

We can also discuss what was raised by Michiel some weeks ago.

redblom:
I mentioned my reservations about using `webdav` as keyword for regular shares to @glpatcern recently at a cs3mesh4eosc wp4.4 call. WebDAV is a protocol that's being used at least in transfers also. How about simply changing that to `share`. 
So: `share`, `webapp`, `datatx`

Also, the receiver of a transfer type share would like to know the transfer size before deciding to accept it (see https://github.com/cs3org/reva/issues/2104). Can we make this a property of the datatx type protocol?

glpatcern:
> I mentioned my reservations about using `webdav` as keyword for regular shares to @glpatcern recently at a cs3mesh4eosc wp4.4 call. WebDAV is a protocol that's being used at least in transfers also. How about simply changing that to `share`. So: `share`, `webapp`, `datatx`

Good point, thought to address that on yet another PR but we could do this here indeed. Now I'd try and find another synonym  given that the OCM endpoint is already called `/shares` and its one JSON parameter including all required properties is called `share`, making it already overused ;-)

> Also, the receiver of a transfer type share would like to know the transfer size before deciding to accept it (see [cs3org/reva#2104](https://github.com/cs3org/reva/issues/2104)). Can we make this a property of the datatx type protocol?

Sure. I can do that if you want.

gmgigi96:
> I mentioned my reservations about using `webdav` as keyword for regular shares to @glpatcern recently at a cs3mesh4eosc wp4.4 call. WebDAV is a protocol that's being used at least in transfers also. How about simply changing that to `share`. So: `share`, `webapp`, `datatx`

I'm not so convinced of renaming the keyword `webdav`. The idea here is that on a share the user knows how to access  (i.e. which protocol to use) the resource. The fact that WebDAV is already used in a transfer also is just a matter of internal implementation of the `datatx` protocol.

redblom:
> 
> > Also, the receiver of a transfer type share would like to know the transfer size before deciding to accept it (see [cs3org/reva#2104](https://github.com/cs3org/reva/issues/2104)). Can we make this a property of the datatx type protocol?
> 
> Sure. I can do that if you want.

Great !

redblom:
> > I mentioned my reservations about using `webdav` as keyword for regular shares to @glpatcern recently at a cs3mesh4eosc wp4.4 call. WebDAV is a protocol that's being used at least in transfers also. How about simply changing that to `share`. So: `share`, `webapp`, `datatx`
> 
> I'm not so convinced of renaming the keyword `webdav`. The idea here is that on a share the user knows how to access (i.e. which protocol to use) the resource. The fact that WebDAV is already used in a transfer also is just a matter of internal implementation of the `datatx` protocol.

My thinking is as follows:
I differentiate between ocm-share type and the protocol to use. Therefor I see the 'properties hierarchy' as follows:
```
   datatx:                  <-- the type
      protocol: "webdav"    <-- the protocol
      srcUri: "https://open-cloud-mesh.org/remote.php/webdav/path/to/resource"
      size: 1000000000000
```
Or the protocol could be taken one level up:
```
   protocol: "webdav"    <-- the protocol
   datatx:               <-- the type
      srcUri: "https://open-cloud-mesh.org/remote.php/webdav/path/to/resource"
      size: 1000000000000
```
But it's all webDAV anyway (for all 3 types I suppose) so that could even be taken for granted (as default protocol) and possibly be left out. So again I tend to lean towards `share`, `webapp`, `datatx`
What do you think?

gmgigi96:
> My thinking is as follows: I differentiate between ocm-share type and the protocol to use. Therefor I see the 'properties hierarchy' as follows:
> 
> ```
>    datatx:                  <-- the type
>       protocol: "webdav"    <-- the protocol
>       sharedSecret: "hfiuhworzwnur98d3wjiwhr"
>       srcUri: "https://open-cloud-mesh.org/remote.php/webdav/path/to/resource"
>       size: 1000000000000
> ```
> 
> Or the protocol could be taken one level up:
> 
> ```
>    protocol: "webdav"    <-- the protocol
>    datatx:               <-- the type
>       sharedSecret: "hfiuhworzwnur98d3wjiwhr"
>       srcUri: "https://open-cloud-mesh.org/remote.php/webdav/path/to/resource"
>       size: 1000000000000
> ```

Probably it was not clear before, and in this PR with @glpatcern we tried to be clearer in the specs, but the `shareType` is the recipient's type, so if a recipient is a user or a group. Currently only the user will be actually implemented as the group one was never discussed/formalized properly since the beginning of the project (the invitation workflow takes into consideration only the discovers of users in fact).

We intended the protocol as the way of accessing/use the share. For example:
- if a user receives a share with protocol `webdav`, the recipient knows that to access the resource he needs to use `webdav` as protocol, and that he can use those resources for browsing, downloading files, create dirs (according to the permissions), but he cannot open a remote app.
- if a user receives a share with protocol `webapp`, he will use the link and make a GET request, to open the file in the remote site with the default app of that resource for example (this might need more formalization).
- if a user receives a share with protocol `datatx`, he knows that he can trigger a data transfer from the remote site to its desired destination and then use the data locally. It's matter of the `datatx` protocol understand which is the way of accessing the resource (in your case webdav, but in theory can be everything).

Considering the last example, can be useful when creating the transfer know which is the source protocol, so we can even put this info in the source link, as `dav+https://open-cloud-mesh.org/remote.php/webdav/path/to/resource`, or add an extra option, like `sourceType`.

> But it's all webDAV anyway (for all 3 types I suppose) so that could even be taken for granted (as default protocol) and possibly be left out.

I'm not so sure about this. I would not strict OCM to have `webdav` as default protocol. For example in the context of web apps this does not make any sense imo.

glpatcern:
> But it's all webDAV anyway (for all 3 types I suppose) so that could even be taken for granted (as default protocol) and possibly be left out. So again I tend to lean towards `share`, `webapp`, `datatx`

Indeed a `webapp` share is a pure https link that is expected to be browser-friendly, and directly usable by the user without any other intermediate service.

Overall it's a bit of an academic question to say that `datatx` is not a "protocol" strictly speaking but a "class" or "type" orthogonal to the protocol, but the reality is also that OCM so far did not have such "type" (the `shareType` refers to users vs groups as mentioned), and IMHO it is not worth to introduce it. Instead, I support the idea to extend the `protocol` to an array and "flatten" this property as one of those protocols, with the potential added value that the same share may have all of the `webdav`, `webapp`, and `datatx` protocols - meaning that it is available to be transferred as well as remotely accessible.

redblom:
@gmgigi96 @glpatcern : OK

glpatcern (approved):

michielbdejong:
> redundant now that the invitation workflow already includes this information

That doesn't hold for the share-with and public-link ("save to your cloud") flows.

glpatcern:
> That doesn't hold for the share-with and public-link ("save to your cloud") flows.

Ah, interesting to know, so those flows don't actually require a prior invitation. ~~Let's discuss the details~~ Edit: just discussed with Hugo, we clearly have to restore those properties indeed.

gmgigi96:
I reverted the recipient display name and as we discussed, I have marked the `providerId` as deprecated for backward compatibility.

spec.yaml
```diff
@@ -210,9 +210,10 @@ definitions:
     required:
       - shareWith
       - name
-      - providerId
+      - resourceId
```
labkode: the providerID should still be part of the message

michielbdejong: Yes, maybe we can mark both providerId and resourceId as required (they would be repeating the same value) during a transition period, and then at the next major version we drop the deprecated providerId.

labkode (commented):
Add comment regarding migration: now is needed both values, later one will be deprecated

gmgigi96:
> Add comment regarding migration: now is needed both values, later one will be deprecated

It's in the description of `providerId`

gmgigi96:
I was just thinking that probably the properties of WebDAV are not fully spelled out. It's not clear what the `sharedSecret` is, if it's basic/digest/bearer authentication. Should we specify the type of authentication?
And in case, the shared secret should change accordingly:
- *basic*: `username:password`
- *digest*/*bearer*: an hash

What do you think? @labkode @glpatcern @michielbdejong

glpatcern:
> I was just thinking that probably the properties of WebDAV are not fully spelled out. It's not clear what the `sharedSecret` is, if it's basic/digest/bearer authentication. Should we specify the type of authentication? And in case, the shared secret should change accordingly:

Good point, but for what I understand the OCM specs are more of a guideline, yet quite detailed, than a full protocol spec like the CS3 APIs (which we use to generate strict bindings). Therefore I don't see this is needed, in the sense that it may remain as it is.

labkode:
@gmgigi96 ideally the protocol should specify how the other end will make a request and I agree that having just `sharedSecret` is not enough as you don't know two things:
- a) How to send them (query parameter? header? body?)
- b) How to encode it: basic auth? jwt?
If we assume every platform implements all possible methods in a) then we need to:
- 1) define what b is in the protocol spec
- 2) or specify in the protocol that implementors should try first one method, then other, ...

glpatcern:
On another side, to complete what we discussed we should rephrase this:

https://github.com/cs3org/OCM-API/pull/57/files#diff-9cfca4a1b73e1e28e30fb9b0b984aad6d4caaf0819c61ed40ad338600531f745R259-R264

And add something like

```
To protect the identity of the parties, for shares created following an OCM invitation,
the user id SHALL be hashed, and recipients implementing the OCM invitation workflow
MAY refuse to process shares coming from unknown parties.
```

And in the invitation request and response we should also detail that we hash the ids (with examples).

gmgigi96:
> @gmgigi96 ideally the protocol should specify how the other end will make a request and I agree that having just `sharedSecret` is not enough as you don't know two things:
> 
>     * a) How to send them (query parameter? header? body?)
> 
>     * b) How to encode it: basic auth? jwt?
>       If we assume every platform implements all possible methods in a) then we need to:
> 
>     * 1. define what b is in the protocol spec
> 
>     * 2. or specify in the protocol that implementors should try first one method, then other, ...

To keep things simple for now we can just specify the authentication type (can be only `basic` and `digest`) as mentioned in the WebDAV RFC https://www.rfc-editor.org/rfc/rfc4918#section-20.1, and maybe expand in the future if we need more.

glpatcern:
> @gmgigi96 ideally the protocol should specify how the other end will make a request and I agree that having just `sharedSecret` is not enough as you don't know two things:

This of course makes even more sense, but my point was in the spirit of general backward compatibility: what is the current behavior of e.g. Nextcloud? and ownCloud? and ...?

In a way, going from something totally unspecified (we barely had the notion of a secret, as an example in a generic `properties` object) to a full spec would be like instantly deprecating implementations that do not follow that spec. What about suggesting a method in the comments?

glpatcern:
OK, following further discussions we have settled in postponing all privacy-related and security-related issues (including e.g. https://github.com/cs3org/OCM-API/issues/23, as well as how WebDAV secrets are expected to be used by implementations) to a topical discussion during the upcoming CS3 conference, when OCM will be one of the prominent topics.

Therefore @gmgigi96 I'd just rephrase `the user id SHALL be hashed` -> `the user id MAY be hashed`, and then this is ready for merge and completes the package of changes to fully support the invitation workflow and the sharing within ScienceMesh.

@michielbdejong could you please review?

spec.yaml
```diff
@@ -262,55 +274,84 @@ definitions:
...
         description: |
-          Share type (user or group share)
+          Recipient share type
         example: user
       resourceType:
         type: string
         description: |
-          Resource type (file, calendar, contact, ...)
-        example: file
-      protocol:
```
michielbdejong, line 277: This looks like a breaking change?
Servers adhering to this spec will not be able to send shares to servers adhering to the previous spec, and vice versa.

michielbdejong: Just like renaming 'providerId' to 'resourceId', renaming 'protocol' to 'protocols' is a complicated project.
We probably need to at least make protocol versions discoverable?

michielbdejong (changes requested):
please clarify whether this PR will break compatibility with existing implementations?

spec.yaml
glpatcern: For the resource type, if `file` "includes" `folder` (which I guess is the case for the existing implementations) then the change is just a clarification. We may well leave it as it was before, that is implicitly saying that folders have type = `file`.

For `protocol` -> `protocols`, this is THE breaking change - and here we already broke compatibility with cs3org/OCM-API#54. I've updated the description of the PR to explicitly state that. What do you mean by making "protocol versions" discoverable?

michielbdejong: I mean since https://github.com/cs3org/OCM-API/pull/37 we have `apiVersion "1.0-proposal1"` in https://cloud.pondersource.com/ocm-provider/ (although I think this is not documented in the spec itself), we should probably tag OCM 2.0 and change that `apiVersion` string to "2.0".
We can create PRs on OC-10 and on NC to make the necessary changes, while keeping backwards compatibility with OCM 1.0.

michielbdejong: Reva doesn't really need to stay compatible with OCM 1.0 because it doesn't support the share-with flow and save-to-your-cloud flow anyway.

glpatcern: I fully support that the `/ocm-provider` endpoint be documented, as in https://github.com/cs3org/OCM-API/issues/51. To be tackled in the near future IMHO.

That said, in principle a save-to-your-cloud flow would be the famous `datatx` protocol (and I wonder how the flow works "now" in e.g. NC and whether after all the `datatx` protocol should build on top of that flow).

For already tagging an `apiVersion "OCM 2.0"`, can we have a quick chat on gitter?

smesterheide:
Hello all, I have worked with OCM in Nexctloud recently. I think all changes are warranted but we should keep in mind that this PR will break OCM compatibility for existing implementations. NC uses the `/ocm-provider` endpoint and reports `apiVersion` as `1.0-proposal1`. It would be great to have this discoverability in the spec itself.

spec.yaml
glpatcern: So we settle this by indeed declaring the `develop` branch as "ready for 2.0", with a clear warning that this is a breaking change compared to OCM 1.0.

We also discussed the `/ocm-provider` endpoint, being on the same page as @smesterheide: the idea is to incorporate `/ocm-provider` in the specs for an upcoming `2.0` tag: I will create a PR for it. It's a bit unfortunate that it's already used as is, whereas it would be great to call it `/.well-known/ocm`, but in https://github.com/cs3org/OCM-API/pull/37#issuecomment-435875108 it seems there are "unpleasant" consequences. Therefore I'll stick to `/ocm-provider`.

glpatcern:
@michielbdejong following the direct discussion, https://github.com/cs3org/OCM-API/pull/57/commits/a59b273e0c694e17877057027fe4fbe7105a9c45 should do. Note that the notification is consistently renamed as well.

spec.yaml
```diff
@@ -235,25 +240,25 @@ definitions:
...
           Cloud Mesh API.
-      providerId:
+      shareId:
         type: string
-        description: >-
-          Identifier to identify the resource at the provider side. This is
-          unique per provider.
+        description: >
+          Identifier to identify the shared resource at the provider side. This is
+          unique per provider and per share.
```
michielbdejong: The phrase "unique per provider" means that each shareId will only be present once at a given provider.
Saying that a share's Id is "unique per share" is a tautology, of course each share has only one ID. :)

michielbdejong: @smesterheide is it also your understanding that this value represents a share (as in, the `id` column of the `oc_share` table) and not a resource?
I think this is especially relevant for the notification endpoint, because the sender may remove the entry from `oc_share` when it receives an unshare notification from the receiver, and for that you need a share id, not a resource id.

michielbdejong: You could make this phrase something like "unique per provider, meaning that if the same resource is shared twice, this shareId will not be repeated."

michielbdejong (approved):
LGTM
Implementers can start supporting the new field name `shareId` alongside its old name `providerId`: send both, and accept whichever one is present.
Just a minor comment about the description of the uniqueness of `shareId`.

glpatcern:
> LGTM
> Implementers can start supporting the new field name `shareId` alongside its old name `providerId`: send both, and accept whichever one is present.

Thanks Michiel, I agree implementers can go ahead using both field names and eventually drop the deprecated one, but it's good to have the specs reflect the final state. I also adapted the phrasing.

PR #59, opened 2023-02-16
Add discovery endpoint

### LINKED ISSUE #51: Endpoint discovery through https://example.com/ocm-provider/

glpatcern:
This PR adds the `/ocm-provider` discovery endpoint, modeled after the already existing implementations.

Also, the OCM API version is pushed to `2.0.0`, given the current state of the `develop` branch. The goal is to agree on an official v2 release of the protocol at the CS3 event in Barcelona (see www.cs3community.org).

Edit: we agreed we would have a dedicated OCM-related meeting, where this and other outstanding items are to be discussed. Out of that meeting (cf. also #69), this PR specifies the behavior of the current implementations, including the exposed version, and adds an optional `capabilities` property.

In a future PR, we will push the version to `1.1.0`.

Closes https://github.com/cs3org/OCM-API/issues/51

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -309,16 +309,16 @@ definitions:
                         are expected to execute the transfer using WebDAV as
                         the wire protocol.
                       example: https://my-cloud-storage.org/remote.php/dav/ocm
-          capabilities:
-            type: array
-            description: |
-              The capabilities exposed at this endpoint according to the present
-              specifications. Implementations MUST expose `/shares` at a minimum.
-            items:
-              type: string
-              enum: ["/shares", "/notifications", "/invite-accepted"]
-            example:
-              ["/shares", "/invite-accepted"]
+      capabilities:
+        type: array
+        description: |
+          The capabilities exposed at this endpoint according to the present
+          specifications. Implementations MUST expose `/shares` at a minimum.
+        items:
+          type: string
+          enum: ["/shares", "/notifications", "/invite-accepted"]
+        example:
+          ["/shares", "/invite-accepted"]
   NewShare:
     type: object
     required:
--- a/spec.yaml
+++ b/spec.yaml
@@ -292,8 +292,8 @@ definitions:
                     a remote shared resource, implementations MAY use this
                     URI as a prefix: e.g. if in a new share request's payload
                     a `protocol.options.sharedSecret` property is defined
-                    (see singleProtocolLegacy example), it may need to be
-                    appended to this URI to access the resource.
+                    (see singleProtocolLegacy example), implementations MAY
+                    need to append it to this URI to access the resource.
                   example: https://my-cloud-storage.org/remote.php/dav/ocm
                 webapp:
                   type: string
@@ -314,7 +314,7 @@ definitions:
       capabilities:
         type: array
         description: |
-          The capabilities exposed at this endpoint according to the present
+          The optional capabilities exposed at this endpoint according to the present
           specifications. As implementations MUST provide `/shares` to be compliant,
           it is not necessary to expose it as a capability.
         items:
--- a/spec.yaml
+++ b/spec.yaml
@@ -245,7 +245,7 @@ definitions:
       apiVersion:
         type: string
         description: The OCM API version this endpoint supports
-        example: 1.0-proposal1
+        example: 1.0.0
       endPoint:
         type: string
         description: The URI of the OCM API available at this endpoint
@@ -288,8 +288,9 @@ definitions:
                         The top-level WebDAV endpoint URI. In order to access a
                         remote shared resource, implementations MAY use this
                         URI as a prefix: e.g. if in a new share request's payload
-                        a `protocol.options.sharedSecret` property is defined,
-                        it should be appended to this URI to access the resource.
+                        a `protocol.options.sharedSecret` property is defined
+                        (see singleProtocolLegacy example), it may need to be
+                        appended to this URI to access the resource.
                       example: https://my-cloud-storage.org/remote.php/dav/ocm
                 - properties:
                     webapp:
--- a/spec.yaml
+++ b/spec.yaml
@@ -41,6 +41,7 @@ paths:
         for more details).
       responses:
         "200":
+          description: The capabilities of this OCM service
           schema:
             $ref: "#/definitions/Discovery"
   /shares:
--- a/spec.yaml
+++ b/spec.yaml
@@ -284,17 +284,30 @@ definitions:
                 - properties:
                     webdav:
                       type: string
-                      description: The top-level WebDAV endpoint
+                      description: |
+                        The top-level WebDAV endpoint URI. In order to access a
+                        remote shared resource, implementations MAY use this
+                        URI as a prefix: e.g. if in a new share request's payload
+                        a `protocol.options.sharedSecret` property is defined,
+                        it should be appended to this URI to access the resource.
                       example: https://my-cloud-storage.org/remote.php/dav/ocm
                 - properties:
                     webapp:
                       type: string
-                      description: The top-level endpoint for web apps
+                      description: |
+                        The top-level endpoint URI for web apps. This value is
+                        provided for documentation purposes, and it SHALL NOT
+                        be intended as a prefix for share requests.
                       example: https://my-cloud-storage.org/external/ocm
                 - properties:
                     datatx:
                       type: string
-                      description: The top-level endpoint for data transfers
+                      description: |
+                        The top-level endpoint URI for data transfers. This value
+                        is provided for documentation purposes, and it SHALL NOT
+                        be intended as a prefix. Furthermore, implementations
+                        are expected to execute the transfer using WebDAV as
+                        the wire protocol.
                       example: https://my-cloud-storage.org/remote.php/dav/ocm
           capabilities:
             type: array
--- a/spec.yaml
+++ b/spec.yaml
@@ -2,7 +2,7 @@ swagger: "2.0"
 info:
   title: Open Cloud Mesh API
   description: Open Cloud Mesh Open API Specification.
-  version: 2.0.0
+  version: 1.0.0
   x-logo:
     url: logo.png
 schemes:
@@ -245,7 +245,7 @@ definitions:
       apiVersion:
         type: string
         description: The OCM API version this endpoint supports
-        example: 2.0.0
+        example: 1.0-proposal1
       endPoint:
         type: string
         description: The URI of the OCM API available at this endpoint
--- a/spec.yaml
+++ b/spec.yaml
@@ -257,7 +257,10 @@ definitions:
         example: MyCloudStorage
       resourceTypes:
         type: array
-        description: A list of all supported resource types with their access protocols
+        description: |
+          A list of all supported resource types with their access protocols.
+          Each resource type is identified by its `name`: the list MUST NOT
+          contain more than one resource type object per given `name`.
         items:
           type: object
           properties:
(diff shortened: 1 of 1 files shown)


glpatcern:
@smesterheide do you mind having a look?

spec.yaml
```diff
@@ -218,6 +231,66 @@ definitions:
...
+        type: boolean
+        description: Whether the OCM service is enabled at this endpoint
+      apiVersion:
+        type: string
+        description: The OCM API version this endpoint supports
+        example: 2.0alpha
+      endpoint:
+        type: string
+        description: The URI of the OCM API available at this endpoint
+        example: https://my-cloud-storage.org/ocm
```
smesterheide: Nextcloud returns the endpoint as `endPoint`. Not sure about others.

spec.yaml
```diff
@@ -218,6 +231,66 @@ definitions:
...
+      provider:
+        type: string
+        description: A friendly branding name of this endpoint
+        example: MyCloudStorage
```
smesterheide, line 257: I noticed this extra property which makes sense and it's not required so it checks out.

smesterheide (commented):
Thanks for taking it upon you to add the /ocm-provider endpoint.

spec.yaml
glpatcern: Good point, same for CERNBox/OCIS and ownCloud as far as I can see. Fixed.

glpatcern:
Thanks Sandro @smesterheide! If also @michielbdejong can have a look, we can merge this ahead of the CS3 conference.

spec.yaml
```diff
@@ -218,6 +231,66 @@ definitions:
...
+            protocols:
+              type: object
+              description: |
+                The supported protocols to access shares at this endpoint
+              anyOf:
+                - properties:
+                    webdav:
+                      type: string
+                      description: The top-level WebDAV endpoint
+                      example: /ocm
```
glpatcern: Actually, reviewing myself, a question that I'd ask to @michielbdejong is what is the purpose of such property (other than declaring that a given protocol is supported), given that shares do have their endpoint anyway and it does not seem that share endpoints are expected to be relative to this one.

Yet I see this is how it was implemented in all EFSSs.

michielbdejong: Interesting! I'll test whether NC & OC-10 still work if you remove this, and if so, then it's probably better to remove it from the spec.

glpatcern: Good! Concerning CERNBox/OCIS, it can be removed (i.e. this can just be specified as an enum)

glpatcern: Speaking with @schiessle, it turns out we may need this prefix. The scenario is what if the EFSS service itself is relocated: what happens to existing shares? If we enforce that shares are relative to some prefix that comes from this discovery endpoint (which we would require clients to regularly refresh), then we have an easy migration path.

To be further discussed...

spec.yaml
```diff
@@ -218,6 +231,66 @@ definitions:
...
+              type: string
+              description: |
+                A supported resource type (file, folder, calendar, contact, ...)
+              example: file
+            shareTypes:
+              type: array
+              description: The supported recipient share types
+              items:
+                type: string
+                enum: ["user", "group"]
```
smesterheide: We may want to add `federation` here as well.

glpatcern:
What we need to add here at this stage is also the list of supported endpoints: funnily enough we discussed this with @schiessle at CS3 Barcelona, and @michielbdejong presented the same thing just minutes after.

Additionally, we need to specify what are the minimal, _required_ features from an EFSS (e.g., sharing with webdav protocol), and clearly mark any other endpoint/option as _optional_.

I plan to address at least those issues prior to marking this ready for review / merge.

glpatcern:
I amended the discovery endpoint with what we agreed today, so it's ready for review. Up to @smesterheide and @michielbdejong whenever you can!

spec.yaml
```diff
@@ -218,6 +231,95 @@ definitions:
...
+              items:
+                type: string
+                enum: ["user", "group", "federation"]
+              example: ["user"]
+            protocols:
+              type: object
+              description: |
+                The supported protocols to access shares at this endpoint.
+                Implementations MUST support `webdav` at a minimum.
+              anyOf:
```
smesterheide: Can we simplify the schema without the use of anyOf?

spec.yaml
```diff
@@ -218,6 +231,95 @@ definitions:
...
+        description: A list of all supported resource types with their access protocols
+        items:
+          type: object
+          properties:
+            name:
+              type: string
+              description: |
+                A supported resource type (file, folder, calendar, contact, ...).
+                Implementations MUST support `file` at a minimum.
+              example: file
```
smesterheide, line 272: Can there be more than one entry for a given resource type (like file) with distinct features (share types, protocol options)?

spec.yaml
```diff
@@ -218,6 +231,95 @@ definitions:
...
+                    webdav:
+                      type: string
+                      description: |
+                        The top-level WebDAV endpoint URI. In order to access a
+                        remote shared resource, implementations MAY use this
+                        URI as a prefix: e.g. if in a new share request's payload
+                        a `protocol.options.sharedSecret` property is defined
+                        (see singleProtocolLegacy example), it may need to be
+                        appended to this URI to access the resource.
+                      example: https://my-cloud-storage.org/remote.php/dav/ocm
```
smesterheide: Are absolute as well as relative URLs allowed?

spec.yaml
```diff
@@ -218,6 +231,95 @@ definitions:
...
+      capabilities:
+        type: array
+        description: |
+          The capabilities exposed at this endpoint according to the present
+          specifications. Implementations MUST expose `/shares` at a minimum.
+        items:
+          type: string
+          enum: ["/shares", "/notifications", "/invite-accepted"]
+        example:
+          ["/shares", "/invite-accepted"]
```
smesterheide: I like the idea! I see this as a way to advertise additional endpoints.

- Should we use method and resource tuples in place of endpoints? Do we gain anything from that?
- Should mandatory endpoints be included as shown?

smesterheide (changes requested):

spec.yaml
glpatcern: How would you suggest to do it then? I understand that the current implementations do have this as a list, or how could we express the ability to support multiple protocols?

spec.yaml
glpatcern: Also here I modeled it from the current implementations, but we could deviate: what would be an example of that?

spec.yaml
glpatcern: I guess we only want absolute URLs here, otherwise they would be relative to "something else" that is defined nowhere. Maybe we should explicitly say that a `top-level endpoint` is an absolute URL.

spec.yaml
glpatcern: Good questions! For the second, true that it does not add any information to expose `/shares` if we state that `/shares` must be implemented, so yeah let's drop it. For the first, could you please provide an example to make sure we understand what would be the model?

spec.yaml
smesterheide: I just wanted to clarify that this is invalid or is it?
```json
{
  "enabled": true,
  "apiVersion": "1.0-proposal1",
  "endPoint": "https://example.com/ocm",
  "resourceTypes": [
    {
      "name": "file",
      "shareTypes": [
        "user"
      ],
      "protocols": {
        "webdav": "/public.php/webdav/"
      }
    },
    {
      "name": "file",
      "shareTypes": [
        "group"
      ],
      "protocols": {
        "webdav": "/public.php/webdav/"
      }
    }
  ]
}
```

spec.yaml
```diff
@@ -30,6 +30,19 @@ parameters:
...
+      description: >
+        This endpoint returns a number of properties used to discover the capabilities
+        offered by a remote cloud storage. The endpoint is named `/ocm-provider` owing
+        to already established practices and constraints with the main cloud storages
+        that implement OCM (see https://github.com/cs3org/OCM-API/pull/37#issuecomment-435875108
+        for more details).
+      responses:
+        "200":
+          schema:
+            $ref: "#/definitions/Discovery"
```
smesterheide:
```suggestion
        "200":
          description: The capabilities of this OCM service
          schema:
            $ref: "#/definitions/Discovery"
```

spec.yaml
smesterheide: This is what I had in mind originally, but it's rather difficult to express. Mixed arrays (oneOf) is supported in OpenAPI 3.0 and tuple validation (prefixItems) requires OpenAPI 3.1.

```
capabilities:
  type: array
  description: |
    The capabilities exposed at this endpoint according to the present
    specifications. As implementations MUST provide `/shares` to be compliant,
    it is not necessary to expose it as a capability.
  items:
    oneOf:
      - type: string
      - type: array
        prefixItems:
          - type: string
            enum: ["GET", "POST", "PUT", "PATCH", "DELETE", "HEAD", "OPTIONS", "TRACE"]
          - type: string
        minItems: 2
        maxItems: 2
  example:
  - /invite-accepted
  - ["POST", "/invite-accepted"]
```

spec.yaml
smesterheide: Would this work?

```
protocols:
  type: object
  properties:
    webdav:
      type: string
      example: https://my-cloud-storage.org/remote.php/dav/ocm
    webapp:
      type: string
      example: https://my-cloud-storage.org/external/ocm
    datatx:
      type: string
      example: https://my-cloud-storage.org/remote.php/dav/ocm
```

spec.yaml
glpatcern: Now I understand, thanks. Indeed I'd argue this is invalid, that is we want `name` to be a unique "key" in the `resourceTypes` map. Not sure we can enforce it with Swagger, I'll check it out - worst case we can still describe the intention in the description and pursue that once we migrate to Open API v3.0 from the draft you started...

spec.yaml
glpatcern: I see, yes, again maybe once we migrate the whole spec to OpenAPI v3.0+? But "do we gain anything with that?" Actually I don't think so. After all e.g. `/invite-accepted` must be a `POST` action by the spec itself, and any other optional endpoint must work using the method according to the spec. In other words I don't see a reason to make the method a "capability" of any given endpoint. In that respect, we could keep it simpler with just the list of endpoints...

spec.yaml
glpatcern: Actually yes, not sure why I expressed it that way in the first place!

spec.yaml
glpatcern: (And on top I think the current implementations assume that that example would be invalid - so I guess we agree on the principle)
It does not seem possible to express this "unicity", also because `resourceType` is an `array`, not a map/dict. To not deviate from current implementations, I'd leave it as an `array` and add this constraint in the description.

glpatcern:
@smesterheide with the current changes we should be close to final, could you please have another look?

spec.yaml
smesterheide: Agreed

spec.yaml
smesterheide: Yep, should be clear

smesterheide (approved):

PR #60, opened 2023-02-20
Add federation share type

smesterheide:
This PR intends to add a new share type named federation. This share type is used over the group share type to indicate to the consumer that the share is simultaneously created at multiple services with the same identifier. The consumer may allow the share to be created first with users joining the federated group later on.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -234,8 +234,8 @@ definitions:
       shareWith:
         type: string
         description: >
-          Consumer specific identifier of the user or group the provider wants
-          to share the resource with. This is known in advance.
+          Consumer specific identifier of the user, group or federation the provider
+          wants to share the resource with. This is known in advance.
           Please note that the consumer service endpoint is known in advance
           as well, so this is no part of the request body.
         example: 51dc30ddc473d43a6011e9ebba6ca770@geant.org
@@ -281,7 +281,7 @@ definitions:
         example: John Doe
       shareType:
         type: string
-        enum: ["user", "group"]
+        enum: ["user", "group", "federation"]
         description: |
           Recipient share type
         example: user


michielbdejong (approved):
Yes! I think this would work. And a very minimal change, great!
I can add support for this in OC-10, along with https://github.com/owncloud/core/pull/40589. Would be good to get some buy-in from the CERNBox team before merging this.

glpatcern (approved):
I like it!

michielbdejong:
@glpatcern do you know who could help add this functionality to OCIS?

glpatcern:
> @glpatcern do you know who could help add this functionality to OCIS?

For the moment it's "us", ownCloud will first have to import the basic OCM implementation in OCIS prior to get here. For the context, OCIS is based on Reva's `edge` branch, which does not implement OCM at all.

michielbdejong:
One side note (this also came up in the [SURF project about federated groups](https://github.com/suRFnet/rd-sram-integration) and I also remember discussing this with @andreasfk in 2021):
Apart from allowing a member of a federation ("federated group" / "virtual organisation") to share a folder or other resource with the other members, it would be cool to be able to attach a sort of "group drive" that doesn't have any particular group member as its owner/sender, but just belongs to the group.
This is relevant for instance when the person who created the group folder leaves the group, the group folder should just stay and not be deleted. Similar to a "network drive" or "shared drive".

smesterheide:
Thanks @michielbdejong and @glpatcern for the swift response!

glpatcern:
As we've seen this today demonstrated at CS3 Barcelona by Nextcloud, I'm taking the liberty to merge it.

PR #61, opened 2023-03-05
Draft text to rewrite this API doc to more of a protocol spec

michielbdejong:


michielbdejong:
I was thinking this document could be more valuable if we add some more text around the different steps sender and receiver go through.

By using "MAY" instead of "SHOULD" in the right places, I think this is possible without adding more requirements or bringing things into scope that are currently out of scope.

For instance, we can document `/ocm-provider` and describe how it is used in practice in some real-world interactions (even in an interoperable way across implementations!), but without making it a requirement. For instance when sharing only from and to the ScienceMesh, this endpoint is not needed.

Another example are the `notificationType` strings that are used interoperably across some (but not all) implementations  in practice; we can at least describe the prevalent `notificationType` strings as things a server "MAY" support, so we don't end up with an eco-system where the same notification type is identified with more than one string.

README.md
```diff
@@ -1,29 +1,82 @@
...
+  * [Share Deletion](#unshare)
+  * [Share Updating](#update)
+  * [Resharing](#reshare)
+
 * [Contributing](#contributing)
 
 ## Scope and assumptions
 
-* Authentication between services is already established. This means that this specification doesn't cover the way a service authenticates incoming API calls (e.g. through an API Key, VPN connection or IP whitelisting). In this scope we assume that the services are already authenticated.
 * Provider knows the consumer (both endpoint and user) when it creates a share with the consumer (also see [#26](https://github.com/cs3org/OCM-API/issues/26)). How this is known is not part of this spec.
```
glpatcern:
```suggestion
* For the core sharing functionality, the provider knows the consumer (both endpoint and user) when it creates a share with the consumer (also see [#26](https://github.com/cs3org/OCM-API/issues/26)). In addition, an optional invitation workflow is available in this specification (see below), which gives the consumer a way to automatically trust a provider (and vice versa). The [ScienceMesh](https://sciencemesh.io) infrastructure provides a managed white list of trusted federated sites.
```

README.md
```diff
@@ -1,29 +1,82 @@
...
 * [Scope and assumptions](#scope-and-assumptions)
-* [API Documentation](#api-documentation)
+* [Specification](#specification)
+  * [Discovery](#discovery)
+  * [Share Creation](#create)
+  * [Share Acceptance](#accept)
+  * [Share Deletion](#unshare)
+  * [Share Updating](#update)
+  * [Resharing](#reshare)
+
```
glpatcern:
```suggestion
 * [Invite](#invite)
 
```

README.md
```diff
@@ -1,29 +1,82 @@
...
+### Share Updating
+TODO: document `"RESHARE_CHANGE_PERMISSION"`
+
+### Resharing
+The `"REQUEST_RESHARE"` and `"RESHARE_UNDO"` notification types may be user by the
+receiving server to persuarde the sending server to share the same resource with another share recipient.
+TODO: document how receiver.com can know if sender.com understood and processed the
+reshare request.
+
```
glpatcern, line 56:
```suggestion
### Invite
If Alice (`alice at sender.com`) and Bob (`bob at receiver.com`) know each other, yet they may not have a mechanism to trust any received share request, as that may be similar to receiving spam.

In this case, Alice may invite Bob to initiate a mutual trust relationship: on the provider side, the `sender.com` service MAY implement an interface for Alice to generate a single-use token and send it to Bob off-band (e.g. via e-mail). In addition, the `sender.com` service MAY integrate this interface with [ScienceMesh](https://sciencemesh.io), and only allow a curated white list of sites as receivers.

On the receiving end, assuming that Bob wishes to accept the invitation, the receiving server MAY provide an interface for Bob to input the received token, or to interact with the ScienceMesh directory. In any case, the receiving server SHOULD make a HTTP POST request to the [/invite-accepted](https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org#/paths/~1invite-accepted/post) endpoint of the sending server, sending the token and disclosing Bob's identity details such as `bob at receiver.com`. If the token matches the one created earlier by Alice, the response MUST include Alice's identity details such as `alice at sender.com`.

Following this step, both services at `sender.com` and `receiver.com` MAY display, respectively, `bob` and `alice` as trusted or white-listed contacts, and enable sharing between them. Sites MAY enforce a policy to only accept shares between such trusted contacts, or MAY display a warning to users when a share from an unknown party is received.

For further details on this concept, see also [#54](https://github.com/cs3org/OCM-API/pull/54) and related issues. For a discussion about trust policies, see [sciencemesh#196](https://github.com/sciencemesh/sciencemesh/issues/196).
```

README.md
```diff
@@ -1,29 +1,82 @@
...
+    {
+      "name":"file",
+      "shareTypes": ["user", "group"],
+      "protocols": {
+        "webdav": (any string)
+      }
+    }
+  ]
+}
+```
```
glpatcern:
```suggestion
When a sending server allows sending to any internet-hosted receiving server, then discovery can happen from the sharee address, using the `/ocm-provider` well-known URL that receiving servers MAY provide according to this [specification](https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org#/paths/~1ocm-provider/get).
```

glpatcern (changes requested):
Hello Michiel, I think we should also resume this in view of v1.1. I have contributed the invite workflow and amended a couple of things also given that the discovery endpoint is officially in the spec. What do you think?

README.md
glpatcern: (here I suggested to remove the example payload because the `/ocm-provider` endpoint is now into the spec)

README.md
```diff
@@ -1,29 +1,82 @@
...
+
+### Share Deletion
+A `"SHARE_ACCEPTED"` notification followed by a `"SHARE_UNSHARED"` notification is
+equivalent to a `"SHARE_DECLINED"` notification.
+
+### Share Updating
+TODO: document `"RESHARE_CHANGE_PERMISSION"`
+
+### Resharing
+The `"REQUEST_RESHARE"` and `"RESHARE_UNDO"` notification types may be user by the
```
glpatcern:
```suggestion
The `"REQUEST_RESHARE"` and `"RESHARE_UNDO"` notification types MAY be used by the
```

README.md
```diff
@@ -1,29 +1,92 @@
...
+receiving server to persuarde the sending server to share the same resource with another share recipient.
+TODO: document how receiver.com can know if sender.com understood and processed the
+reshare request.
+### Invite
```
glpatcern, line 57:
```suggestion

### Invite
```

glpatcern:
I have incorporated the `invite` description so it's easier to read through. For the rest, I leave it to you to have a look / improve.

README.md
```diff
@@ -1,29 +1,93 @@
...
 * [Scope and assumptions](#scope-and-assumptions)
-* [API Documentation](#api-documentation)
+* [Specification](#specification)
+  * [Discovery](#discovery)
+  * [Share Creation](#create)
+  * [Share Acceptance](#accept)
+  * [Share Deletion](#unshare)
+  * [Share Updating](#update)
+  * [Resharing](#reshare)
+ * [Invite](#invite)
```
glpatcern:
```suggestion
  * [Invite](#invite)
```

README.md
```diff
@@ -1,29 +1,93 @@
...
+
+Following this step, both services at `sender.com` and `receiver.com` MAY display, respectively, `bob` and `alice` as trusted or white-listed contacts, and enable sharing between them. Sites MAY enforce a policy to only accept shares between such trusted contacts, or MAY display a warning to users when a share from an unknown party is received.
+
+For further details on this concept, see also [#54](https://github.com/cs3org/OCM-API/pull/54) and related issues. For a discussion about trust policies, see [sciencemesh#196](https://github.com/sciencemesh/sciencemesh/issues/196).
+## Contributing
```
glpatcern, line 69:
```suggestion


## Contributing
```

glpatcern (approved):
I suggest to move on with all this, as discussed we can always improve and the added text already documents several undocumented aspects. I will also build on top of this to document the webdav access in OCM v1.0, as discussed in #70.

PR #65, opened 2023-03-10
Clarify versions

### LINKED ISSUE #62: "protocol" or "protocols", which should implementers use?
There is now a difference between the Reva implementation and the Nextcloud / Seafile / OC-10 implementation of OCM:
Reva uses "protocols" and all the others use "protocol".

Our main spec page https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org should describe that reality more clearly.

@glpatcern it's probably best if we have a video call about this to discuss how we want to deal with this nuance?

michielbdejong:
Fixes #62
See https://github.com/cs3org/OCM-API/blob/fix-62/README.md#api-documentation for a preview

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,8 @@ This repository contains the [OpenAPI](https://github.com/OAI/OpenAPI-Specificat
 
 The specification can be rendered as HTML documentation using [ReDoc](https://github.com/Redocly/redoc):
 
-* [API Reference Documentation](https://cs3org.github.io/OCM-API/docs.html)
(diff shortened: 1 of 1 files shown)


glpatcern:
Agreed that we need to clarify this, but IMHO we are close to have ScienceMesh work with "Network beta". So before committing this I'd rather fix what remains to be fixed, and then phrase it like:
```
version 1.0: current official and supported version
develop branch: future version supported by ScienceMesh
```

michielbdejong:
Updated, thanks!

glpatcern (approved):

Issue #68, opened 2023-03-20
Backwards compatibility


smesterheide:
I am not sure if it was decided yet whether the next version (v2.0.0 as it stands now) will be backwards compatible with v1.0.0. As a reminder v1.0.0 is the only spec available with some degree of interoperability among vendors. As this is the stated goal of OCM we should maybe discuss this aspect before settling on the next version. As a starter maybe we can first agree on the notion of backwards compatibility.

- Should the specification itself include all endpoints, schemas, etc. from previous versions (marked as deprecated)?
- Should the API version be encoded in the URL path, eg. `{server}/v1`
- How can an implementation advertise compatibility with multiple versions?
- How does the discoverability endpoint `/ocm-provider` tie into that?
- Do we maybe need a minor release first without any new features to facilitate backwards compatibility with new API versions?

glpatcern:
Let's discuss this as part of the call next Wednesday. I have a [draft spec](https://github.com/glpatcern/OCM-API/tree/reconcile) where I keep backwards compatibility whilst still supporting the new endpoints, and IMHO part of the questions are easily answered. But as usual the details are worth being discussed (e.g. how does the discoverability endpoint `/ocm-provider` tie into all this).

glpatcern:
I'll try and answer the points above from the discussion we had yesterday, following the decision to go for version 1.1.0, but I have further questions:

* Should the specification itself include all endpoints, schemas, etc. from previous versions (marked as deprecated)?
  * Yes. There's a single field marked as deprecated in #69
* Should the API version be encoded in the URL path, eg. `{server}/v1`?
  * I don't think this is worth it at this point.
* How can an implementation advertise compatibility with multiple versions? How does the discoverability endpoint /ocm-provider tie into that?
  * The discovery endpoint (#59) provides the supported `protocols`. Can we assume that IF the list of supported protocols includes more than `webdav`, THEN the implementation must be v1.1.0 capable? I think yes, though this was not explicitly stated at the meeting. Then, whether a v1.1-capable implementation is compatible with a v1.0 implementation: what about adding this in the discovery endpoint as an extra capability of some form?

smesterheide:
> Let's discuss this as part of the call next Wednesday. I have a [draft spec](https://github.com/glpatcern/OCM-API/tree/reconcile) where I keep backwards compatibility whilst still supporting the new endpoints

[...]

PR #69, opened 2023-03-24
Renamed/reintroduced properties to ease backwards compatibility

### LINKED ISSUE #66: Do we want to support more than one protocol at a time?
This discussion started on another issue, but it is important to settle on it.

Following #54 and #57, we have introduced the option to specify more than a protocol in a share action.

The rationale of this is to enable richer use-cases such as [what I demonstrated at CS3 Barcelona](https://indico.cern.ch/event/1210538/contributions/5206694), where the user could access the data (via `webdav`) and the remote application _at the same time_ (via `webapp`), or what @schiessle came up with as another example: imagine you share a calendar event, and you want the recipient to access the event data via CalDAV (i.e. protocol = `webdav`), as well as connecting to some groupware tool such as Zoom, Nextcloud talk, etc. (i.e. protocol = `webapp`).

From @smesterheide:

> I think the [original definition](https://cs3org.github.io/OCM-API/docs.html?branch=v1.0.0&repo=OCM-API&user=cs3org) was superior to having different protocols defined as properties. Can we use name as a [discriminator](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/)?

The original definition was just an object, which does not offer any clue as of how implementations are supposed to encode the properties. No surprise that compatibility across different vendors was not ideal (e.g. only recently did Seafile announce compatibility with Nextcloud).

Using just a single property - the `shareType` rather than the name I guess - as discriminator implies that there's a single "dimension" to express all options. This has two issues:
1. We'd need to come up with counterintuitive labels to represent the whole `shareType` x `protocol` matrix: for the above examples, something like `file-webdav`, `file-app`, `caldav`, `calendar-talk-app`?
2. Even worse, a user that wants to share a file with two protocols / access methods _at the same time_, would need to create two independent shares, with poor UX (why would I need to share twice?) and inconsistent scenarios thereafter - what if one of the two shares is later removed? does the recipient keep the permissions on the other one? either way it's wrong/inconsistent.

More opinions welcome!

glpatcern:
This closes #66 and #68, ~~to be discussed~~ according to the discussion we had today 29/03/2023.

In particular:
* The spec is backwards compatible, and descriptions were added to clarify how to use the additional properties vs the legacy ones.
* In `/shares`, the `protocol.name` serves as switch between legacy and new format, and the `protocol.options` field has been marked as deprecated, in favor of `protocol.webdav` and similar fields as already defined.
* In the new protocol properties, the permissions are now fully spelled out, as opposed to relying on an external permissions namespace.

https://cs3org.github.io/OCM-API/docs.html?branch=reconcile&repo=OCM-API&user=glpatcern for a rendered preview.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -452,7 +452,9 @@ definitions:
                   description: |
                     An URI to access the remote resource. The URI MAY be relative,
                     in which case the prefix exposed by the `/ocm-provider` endpoint MUST
-                    be used, or it may be absolute (recommended).
+                    be used, or it may be absolute (recommended). Additionally, the URI
+                    MAY include a secret hash in the path, in which case there MAY be
+                    no associated `sharedSecret`.
             webapp:
               type: object
               properties:
@@ -460,7 +462,10 @@ definitions:
                   type: string
                   description: |
                     A templated URI to a client-browsable view of the shared resource,
-                    such that users may use the web applications available at the site
+                    such that users may use the web applications available at the site.
+                    The URI MAY include a secret hash in the path. If the path includes
+                    a `{relative-path-to-shared-resource}` placeholder, implementations
+                    MAY replace it with the actual path to ease user interaction.
                 viewMode:
                   type: string
                   description: |
@@ -487,7 +492,8 @@ definitions:
                   description: |
                     An URI to access the remote resource. The URI MAY be relative,
                     in which case the prefix exposed by the `/ocm-provider` endpoint MUST
-                    be used, or it may be absolute (recommended).
+                    be used, or it may be absolute (recommended). Additionally, the
+                    URI MAY include a secret hash in the path.
                 size:
                   type: integer
         example:
@@ -499,9 +505,8 @@ definitions:
           singleProtocolNew:
             name: "multi"
             webdav:
-              sharedSecret: "hfiuhworzwnur98d3wjiwhr"
               permissions: ["read"]
-              uri: "https://open-cloud-mesh.org/remote.php/webdav/path/to/spec.yaml"
+              uri: "https://open-cloud-mesh.org/remote.php/webdav/share-hash/path/to/spec.yaml"
           multipleProtocols:
             name: "multi"
             options:
--- a/spec.yaml
+++ b/spec.yaml
@@ -337,7 +337,7 @@ definitions:
     properties:
       shareWith:
         type: string
-        description: >
+        description: |
           Consumer specific identifier of the user, group or federation the provider
           wants to share the resource with. This is known in advance.
           Please note that the consumer service endpoint is known in advance
@@ -350,12 +350,12 @@ definitions:
       description:
         type: string
         description: Optional description of the resource (file or folder).
-        example: >-
+        example: |
           This is the Open API Specification file (in YAML format) of the Open
           Cloud Mesh API.
       providerId:
         type: string
-        description: >
+        description: |
           Identifier to identify the shared resource at the provider side. This is
           unique per provider such that if the same resource is shared twice,
           this providerId will not be repeated.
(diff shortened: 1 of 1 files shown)


glpatcern:
@smesterheide looking forward to hear from the permissions namespace used by Nextcloud, and whether we can pick from it to further detail the spec. Apart that, I believe I have covered all we discussed, and I actually recommend to compare this PR with the `v1` tag to see how backwards compatibility is addressed:

https://github.com/cs3org/OCM-API/compare/v1...glpatcern:OCM-API:reconcile

glpatcern:
@schiessle I guess you shall get an update from @smesterheide, but please have a look at this - once merged, we can target a `v1.1` that is backwards compatible with the current `v1.0`, and Nextcloud can start implementing multi-protocol support as well as the other new features with a simple switch on `protocol.name`.

I appreciate that we also got SeaFile onboard with @killing, and @dragotin and ownCloud have shown interest in porting the OCM implementation developed at CERN to OCIS. Hopefully, we can keep driving OCM forward as a collective effort from all involved parties.

smesterheide:
I found this diff to be useful: https://github.com/cs3org/OCM-API/compare/v1.0.0...glpatcern:OCM-API:reconcile

glpatcern:
> I found this diff to be useful: [v1.0.0...glpatcern:OCM-API:reconcile](https://github.com/cs3org/OCM-API/compare/v1.0.0...glpatcern:OCM-API:reconcile)

Even better than my diff. @smesterheide what about the `"{http://open-cloud-mesh.org/ns}share-permissions"` example? If you can find any reference that can be included, I think we can leave it in the example for the options. Otherwise I'd just leave it out as it stands now.

smesterheide:
I see no problems with this PR and it provides some good improvements. It restores all v1 properties while bringing support for multi protocol shares. Here is the detailed list of changes I have found (compared to v1):
* invite-accepted endpoint (no change to current develop)
* sender is now required (little impact)
* providerId: its meaning was clarified (it's the id of the shared resource)
* shareType is an enum which now includes federation
* resourceType mentions folder now (is subsumed as file in Nextcloud)
* expiration added (optional)
* protocol
  * options is no longer required.
  * name includes multi now which when used opens up the possibility to define protocol options for one or more protocols. These protocol options are added as additional properies (webdav, webapp, datatx) on the  protocol object.
  * ~~name and options are no longer defined under properties (mistake or is this how additionalProperties works?)~~

smesterheide:
> Even better than my diff. @smesterheide what about the `"{http://open-cloud-mesh.org/ns}share-permissions"` example? If you can find any reference that can be included, I think we can leave it in the example for the options. Otherwise I'd just leave it out as it stands now.

I am still looking into that. I think we can discuss this on a different occasion

glpatcern:
>       * name and options are no longer defined under properties (mistake or is this how additionalProperties works?)

It was my mistake, I fixed it. Otherwise thanks for the review, you gave a fair summary of what future v1.1 brings on top of v1.0.

smesterheide:
By email I asked if we could include the protocol *example* which includes a webdav based share with the legacy format: 

```
"protocol": {
  "name":"webdav",
  "options":{
      sharedSecret: "hfiuhworzwnur98d3wjiwhr",
      permissions: "{http://open-cloud-mesh.org/ns}share-permissions"
  }
}
```

Thanks @glpatcern for already adding it back to our prospective v1.1

spec.yaml
```diff
@@ -393,72 +393,134 @@ definitions:
...
-            viewMode: "write"
-          datatx:
-            sharedSecret: secret
-            srcUri: "https://open-cloud-mesh.org/remote.php/webdav/path/to/resource"
-            size: 1000000000000
+          singleProtocolLegacy:
+            name: "webdav"
+            options:
+              sharedSecret: "hfiuhworzwnur98d3wjiwhr"
+              permissions: "some permissions scheme"
```
glpatcern, line 504: @smesterheide any news about the permissions scheme used in Nextcloud? If not, I propose to leave this in the example as a generic placeholder, it would fit the current as well as future implementations. In this case, can you please approve the PR?

labkode (approved):

michielbdejong (approved):
LGTM

spec.yaml
glpatcern: As discussed live, we keep it like that

PR #74, opened 2023-05-12
Add warning for client implementers

michielbdejong:
From discussion with @yasharpm

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -424,7 +424,7 @@ definitions:
                 its options MAY be given in the opaque `options` payload for
                 compatibility with v1.0 implementations (see examples).
                 CAREFUL: client implementers should be aware that v1.1 servers
-                will support both `webdav` and `multi`, but v1.0 servers will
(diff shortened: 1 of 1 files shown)


spec.yaml
```diff
@@ -423,6 +423,9 @@ definitions:
                 Otherwise, at least `webdav` is expected to be supported, and
                 its options MAY be given in the opaque `options` payload for
                 compatibility with v1.0 implementations (see examples).
+                CAREFUL: client implementers should be aware that v1.1 servers
+                will support both `webdav` and `multi`, but v1.0 servers will
```
```suggestion
                MAY support both `webdav` and `multi`, but v1.0 servers MAY
```
Am I too conservative? A v1.1 server could e.g. support the invite workflow but only implement webdav sharing with the "legacy" format. It is only if multi-protocols are advertised on `/ocm-provider` that a v1.1 server could use `multi`.

smesterheide (approved):

PR #76, opened 2023-06-13
Fixed paths used in the discovery endpoint

### LINKED ISSUE #70: Document current translation that happens for webdav
I'm a bit embarrassed that I don't know this by heart but I'm just experimentally observing how to access a webdav share from say nc1.docker. I see my nc2.docker make requests like:

`PROPFIND` https://nc1.docker/remote.php/dav/files/einstein/asdf/qwer/asdf
And with a `'requesttoken' => '2oCY+AwkE9BKYBynQflAS+G4RnQcU3yyHgkewQblqNw=:q6v2ikhhSZR+I1aTcL0kGNjgCxUrfBSHcH5Q8ESPza8='` header (I'm still figuring out if this is what carries the sharedSecret and if so, how this is put into what looks like two base64-encoded strings).

I see https://nc1.docker/ocm-provider advertises https://nc1.docker/remote.php/webdav/ as the webdav root so I think that was then redirected to https://nc1.docker/remote.php/dav/

I'll get to the bottom of this so we can add it to the spec.

glpatcern:
This follows a reverse engineering session to expose a share from CERNBox to ownCloud 10 and Nextcloud 26. Closes #70.

- [x] OC10 access works
~~- [ ] NC26 access works~~

Edit: Nextcloud has been proven to follow the OCM standard when sending a share to a remote system, but to **fully disregard it** when accessing a remote system. Details in the comment below and in https://github.com/pondersource/nc-sciencemesh/issues/373. Therefore, the amended text just reflects how ownCloud 10 accesses remote OCM shares, that essentially boils down to exposing paths and not full URLs in the `/ocm-provider` discovery endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,16 +23,15 @@ the [OpenAPI](https://github.com/OAI/OpenAPI-Specification) (fka Swagger) specif
 * For the core sharing functionality, the provider knows the consumer (both endpoint and user) when it creates a share with the consumer (also see [#26](https://github.com/cs3org/OCM-API/issues/26)). In addition, an optional invitation workflow is available in this specification (see below), which gives the consumer a way to automatically trust a provider (and vice versa). The [ScienceMesh](https://sciencemesh.io) infrastructure provides a managed white list of trusted federated sites.
 * Consumer doesn't have to accept a share, the resource will be available to the consumer immediately ([#25](https://github.com/cs3org/OCM-API/issues/25)).
 * Dealing with incoming shares is a vendor specific implementation. One vendor might use an 'accept before' process while another vendor might use a 'decline after' approach. This is considered part of the UX and thus not part of the interaction between different vendors. However, the consumer could notify the provider by using the introduced `/notifications` endpoint (also see [#27](https://github.com/cs3org/OCM-API/issues/27)).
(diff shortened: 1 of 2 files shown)


glpatcern:
To keep a summary record of the investigations concerning Nextcloud, this is how it accesses a remote OCM share.

To be noted that part of this logic appears to come from the [Open Collaboration Services](https://www.freedesktop.org/wiki/Specifications/open-collaboration-services) standard (see #70), part is Nextcloud's own logic.

1) Check the remote `/ocm-provider` endpoint and ensure that `apiVersion` = `1.0-proposal1`. Otherwise, bail out.
2) Send an OCM notification that the share has been accepted, and ensure it return `201 CREATED`. Otherwise, bail out.
3) Check the remote `/ocs-provider` endpoint. Possibly take `services.FEDERATED_SHARING.endpoints.webdav` as `<path>` for later accessing (not confirmed).
4) Check that `/status.php` returns HTTP 200. This endpoint is exposed by ownCloud and Nextcloud, and the check is to ensure that the remote end is either ownCloud or Nextcloud. If not, bail out.

Nextcloud may eventually call `POST /index.php/apps/files_sharing/shareinfo`, passing as `t` query parameter the sharedSecret of the share. This is expected to give info about the remote share within a pure Nextcloud federation, and it's a fallback mechanism when one of the previous checks failed. But it is not called when accessing a ownCloud 10 share.

michielbdejong:
I can confirm point 1, [my ocm-stub serves](https://github.com/michielbdejong/ocm-stub/blob/fad3ceb/stub.js#L150) the exact string `1.0-proposal1` as its OCM version and you're correct to note that using semver here would be better.

But from memory I can't confirm points 2, 3 and 4, would be nice to create explicit tests for this in the ocm test suite!

michielbdejong (approved):
LGTM

Issue #31, opened 2017-01-30
Notifications endpoint might also return a 202 accepted


dvh:
At the moment we assume that the only valid success response is a `201 Created`. However, if the receiver wants to process the notification in another way, it might also return a `202 Accepted` response to inform the sender that the notification has been received but it's not possible for the receiver to fetch the notification later, because the receiver implemented it's own notification handler.

michielbdejong:
The side-effects of notifications have always been a slightly confusing part of OCM. I interpret the word 'notification' to mean that no report of completion of the actions it triggers is necessary. In fact, we consider those triggered actions as 'side effects' and not as 'completion of the request'.

For example if I notify you "your house is on fire" you may respond "201 thanks for the heads-up!", that's a notification.
But if I send you "please extinguish the fire in your house" then that is not a notification but an API request, and you may respond "202 accepted".

With that analysis, and as @glpatcern mentioned in #77  this would be a breaking change, let's keep things as they are for now, closing.

michielbdejong:
(see also [discussion of notifications as a general collection for "anything you might do"](https://github.com/cs3org/OCM-API/issues/34#issuecomment-281103310))

PR #77, opened 2023-06-19
Improved spec for notifications

### LINKED ISSUE #46: Are notifications allowed to have side-effects?
A little philosophical question I ran into while writing the test suite:

When Alice shares a resource with Bob, she cannot share it again. If she tries to share it a second time, her GUI will warn her 'Bob already has access'. But if Bob sent back a notification like `SHARE_DECLINED` or `SHARE_REMOVED` the that will in practice usually have the side-effect that Alice can share the resource with Bob again.

This suggests that either sharing with someone should be idempotent, or the notifications and their side-effects should be standardised.

glpatcern:
This PR clarifies how notifications are actually used by implementations, in particular Nextcloud.

- Added a full example and details of possible notification types
- Removed a `required` but unspecified field
- Added `providerId` (already present in Nextcloud's payload) as a `required` field. Without it the notification cannot be consumed by the target!
- Extended the description to cover for the expected/typical side effects as MAY behaviors following a notification

To be noted that the response HTTP code in case of success was/is `201 CREATED`. It is debatable why such a response was chosen and whether a standard `200 OK` or even `202 ACCEPTED` (cf https://github.com/cs3org/OCM-API/issues/31) would be more appropriate, but this is left outside this PR as it would be a breaking change.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -116,8 +116,10 @@ paths:
         other party about a change about a previously known entity, such as a share or
         a trusted user.
         For example, a notification MAY be sent by a recipient to let the provider know
-        that the recipient declined a share. Similarly, it MAY be sent by a provider
-        to let the recipient know that the provider removed a given share.
+        that the recipient declined a share. In this case, the provider site MAY mark the
+        share as declined for its user(s).
+        Similarly, it MAY be sent by a provider to let the recipient know that the provider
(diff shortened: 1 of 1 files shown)


glpatcern:
I'm moving forward and merging this to avoid leaving stuff behind. As explained, this is just adding comments and fixing the spec according to how it was actually implemented.

PR #79, opened 2024-01-12
Add MFA capability and permission

mickenordin:
This PR is a proposal to add a capability `/mfa-enforcing` as well as a permission `mfa-enforced` to the specification. The version of the specification is bumped to 1.2.0.

If an OCM provider has the capability `/mfa-enforcing` it will respond with a boolean on the endpoint /mfa-enforcing to indicate whether or not it will try to comply with a MFA requirement set as a permission on a share. If the sharer OCM provider trusts the sharee OCM provider the sharer MAY set the permission `mfa-enforced` on a share.

A complient OCM provider that signals mfa-enforcing `true` MUST not allow access to a resource to a user that has not provided a second factor to establish the identity of the user with greater confidence.

Since there is no way to guarantee that the sharee OCM provider will actually enforce the MFA requirement, it is up to the sharer OCM provider to establish a trust with the OCM sharee provider such that it is reasonable to assume that the sharee OCM provider will honor the MFA requirement. This establishment of trust will inevitably be implementation dependent, and can be done for example using a pre approved allow list of trusted OCM providers. The procedure of establishing trust is out of scope for this specification.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -108,21 +108,31 @@ paths:
               type: string
           schema:
             $ref: "#/definitions/Error"
-  /mfa-capable:
+  /mfa-enabled:
     get:
-      summary: Inform the sender that the provider will enforce MFA requirements.
+      summary: Request whether the consumer of a share has been authenticated with MFA.
       description: >
-        Signal that this OCM provider has the capability to enforce MFA requirements.
-        A sender MAY set the permission `mfa-enforced` on a share to this provider.
-        NOTE: There is no guarantee that the provider will actually enforce any MFA
+        When a consumer attempts to access a share that was created with the `mfa-enabled`
+        permission, the provider MUST request to the consumer's server whether the consumer
+        has been authenticated with MFA. The consumer server SHOULD enforce that a session
+        exists for the consumer and that it has been authenticated with MFA, or prompt
+        the consumer in order to elevate their session to MFA.
+        Note: There is no guarantee that the consumer service actually enforces any MFA
         requirements, so a trust must be established before relying on this capability.
+      parameters:
+        sharedSecret:
+          type: string
+          description: |
+            The secret identifying the share being accessed.
       responses:
         "200":
           description: |
-            The OCM service claims that it is capable of enforcing MFA requirements.
+            The user accessing the resource via the given `sharedSecret` has been
+            authenticated with MFA.
         "404":
           description: |
-            The OCM service does not have the capability to enforce MFA requirements.
+            There is no user accessing a share with the given `sharedSecret`, or
+            the endpoint does not have the capability to enforce MFA requirements.
   /notifications:
     post:
       summary: Send a notification to a remote party about a previously known entity
@@ -336,7 +346,7 @@ definitions:
           it is not necessary to expose it as a capability.
         items:
           type: string
-          enum: ["/notifications", "/invite-accepted", "/mfa-capable"]
+          enum: ["/notifications", "/invite-accepted", "/mfa-enabled"]
         example:
           ["/invite-accepted"]
   NewShare:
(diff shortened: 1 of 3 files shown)


spec.yaml
```diff
@@ -2,7 +2,7 @@ swagger: "2.0"
 info:
   title: Open Cloud Mesh API
   description: Open Cloud Mesh Open API Specification.
-  version: 1.1.0
+  version: 1.2.0
```
glpatcern: For now I'd keep the version as is, we shall see when to tag a 1.2.0 version - which would include other features as well - in the future. Maybe the CS3 workshop can be a "good" moment.

spec.yaml
```diff
@@ -108,6 +108,19 @@ paths:
               type: string
           schema:
             $ref: "#/definitions/Error"
+  /mfa-enforcing:
```
glpatcern: Maybe this endpoint could be better named `/mfa-capable`? What I mean is that an `mfa-capable` OCM provider does not necessarily enforce MFA in all shares, but only on those where the `mfa-enforced` permission is set.

Also, I think it can be simplified with just an empty `200` response, as an implementation that does not have such capability would respond with a `404`.

glpatcern (commented):
I like the approach, including the additional capability a provider is expected to implement.

I think part of the description in this PR could be fed into the `README.md`. Otherwise, I left a couple of comments.

spec.yaml
mickenordin: Good point!

spec.yaml
mickenordin: Fixed in the latest commit

glpatcern (approved):
Looks good! @smesterheide and/or @michielbdejong any further thoughts?

michielbdejong (commented):
We should maybe have a more generic way to discover capabilities.
In practice, current implementations rely on `/ocm-provider` I think.
Did we want to move away from that?
In any case, shouldn't we use something like `/.well-known/ocm-configuration`?

glpatcern:
I remember that `/.well-known` was/is problematic for Nextcloud, and they explicitly asked to stick to the current `/ocm-provider` for discovery. Perhaps worth checking this again now...

mickenordin:
Does it look like you expected now @glpatcern? I am not sure I did the right thing when merging your changes...

glpatcern:
Yes, I think this is OK (modulo one spurious thing easy to fix). For the discovery endpoint, that's something that was discussed in https://github.com/cs3org/OCM-API/pull/37#issuecomment-435875108 and at the time it was decided NOT to go to `.well-known`. Time to rediscuss it, but outside of this PR.

spec.yaml
```diff
@@ -514,14 +540,14 @@ definitions:
               sharedSecret: "hfiuhworzwnur98d3wjiwhr"
             webdav:
               sharedSecret: "hfiuhworzwnur98d3wjiwhr"
-              permissions: ["read"]
-              uri: "https://open-cloud-mesh.org/remote/dav/ocm/hfiuhworzwnur98d3wjiwhr/path/to/resource.txt"
+              permissions: ["read", "write"]
+              uri: "https://open-cloud-mesh.org/remote.php/webdav/share-secret/path/to/resource.txt"
```
```suggestion
              uri: "https://open-cloud-mesh.org/remote/dav/ocm/hfiuhworzwnur98d3wjiwhr/path/to/resource.txt"
```

spec.yaml
```diff
@@ -514,14 +540,14 @@ definitions:
...
           multipleProtocols:
             name: "multi"
             options:
             webdav:
-              permissions: ["read"]
-              uri: "https://open-cloud-mesh.org/remote/dav/ocm/hfiuhworzwnur98d3wjiwhr/path/to/resource.txt"
+              permissions: ["read", "mfa-enforced"]
+              uri: "https://open-cloud-mesh.org/remote.php/webdav/share-secret/path/to/resource.txt"
```
```suggestion
              uri: "https://open-cloud-mesh.org/remote/dav/ocm/hfiuhworzwnur98d3wjiwhr/path/to/resource.txt"
```

glpatcern (approved):

PR #82, opened 2024-03-21
Introduce a .well-known discovery endpoint and DNS SRV records

### LINKED ISSUE #38: Server alias
In order to improve the usability, it might be useful to provide a way to set aliases for a service, allowing users to share using that alias instead of the full url of the service.

Imagine the case of CERNBox: the service is deployed at `cernbox.cern.ch`, and this should be the endpoint used for auto discovery/getting information about the service (we're not planning to put this information in the top domain, which is not controlled by us). In the current model, a user would have to put `username_or_email@cernbox.cern.ch` to share something, which is not intuitive. If we set "cern.ch" as an alias, then it would be possible to share with `username_or_email@cern.ch`- which is generally the way users know each other -, and the client would contact `cernbox.cern.ch`. 

All of this is assuming that the services are configured manually, since - as I understood - you, as a sysadmin, need to approve it first.

glpatcern:
Following today's discussions, would this be enough? Maybe complemented with some details in the `README.md` file?

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -30,15 +30,28 @@ parameters:
       the HAL navigation links (e.g. `_links.next.href`) to paginate. These
       links enable the possibility to use vendor specific pagination.
 paths:
-  /ocm-provider:
+  /.well-known/ocm:
     get:
       summary: Discovery endpoint
       description: >
-        This endpoint returns a number of properties used to discover the capabilities
-        offered by a remote cloud storage. The endpoint is named `/ocm-provider` owing
-        to already established practices and constraints with the main cloud storages
-        that implement OCM (see https://github.com/cs3org/OCM-API/pull/37#issuecomment-435875108
-        for more details).
+        Following RFC 8615, this endpoint returns the properties and capabilities
+        offered by a remote cloud storage. This endpoint is to be served as a
+        top-level URL, e.g. as in `https://my-cloud-storage.org/.well-known/ocm`.
+      responses:
+        "200":
+          description: The capabilities of this OCM service
+          schema:
+            $ref: "#/definitions/Discovery"
+  /ocm-provider:
+    get:
+      summary: Legacy discovery endpoint
+      description: >
+        This endpoint returns the properties and capabilities offered by a remote
+        cloud storage. It is a replica of `/.well-known/ocm` and it is required
+        to be supported, owing to previously established practices by the main
+        cloud storages that have implemented OCM (for more details, see
+        https://github.com/cs3org/OCM-API/pull/37#issuecomment-435875108).
+        This endpoint is to be served as a top-level URL.
       responses:
         "200":
           description: The capabilities of this OCM service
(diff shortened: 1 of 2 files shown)


spec.yaml
```diff
@@ -30,13 +30,24 @@ parameters:
       the HAL navigation links (e.g. `_links.next.href`) to paginate. These
       links enable the possibility to use vendor specific pagination.
 paths:
-  /ocm-provider:
+  /.well-known:
```
mickenordin: This should probably be /.well-known/ocm-provider or something: https://en.wikipedia.org/wiki/Well-known_URI

glpatcern: That's right, when I drafted it I just forgot that `/.well-known` lives at the top level and we need a sub path.

By the way, the spec was unclear about the discovery endpoint being served as a top-level URL, therefore I added a specific line about that.

mickenordin (approved):
Looks good!

DanScharon:
wonderful! 
Is this intended to enable the discovery of a user's federated instance? 
For example I would enter the email address of a person I wish to share with (person1@example.com) and my instance would look up https://example.com/.well-known/ocm which would point to https://cloud-storage.example.com and my instance would ask https://cloud-storage.example.com whether it has a user with the email address person1@example.com that I could share with?

glpatcern:
> wonderful! Is this intended to enable the discovery of a user's federated instance?

@DanScharon that's precisely the idea, assuming that the vendors will implement that mechanism in Nextcloud, OCIS, etc.

I intend to include that workflow in the `README.md`, possibly coupled with the other idea that emerged at the CS3 workshop, that is the DNS SRV record.

DanScharon:
> > wonderful! Is this intended to enable the discovery of a user's federated instance?
> 
> @DanScharon that's precisely the idea, assuming that the vendors will implement that mechanism in Nextcloud, OCIS, etc.
> 
> I intend to include that workflow in the `README.md`, possibly coupled with the other idea that emerged at the CS3 workshop, that is the DNS SRV record.

Excellent! Thank you for including this in the spec!

glpatcern:
I'd love a second read of the paragraph I added to the `README.md`, @mickenordin would you please have a look? Thanks!

mickenordin (approved):
I think this is a fine addition.

We should probably extract relevent information from this discussion and add it to the spec text it self: https://github.com/cs3org/OCM-API/pull/37#issuecomment-435875108

But I dont think it needs to happen in this pr, we can do it separately.

glpatcern:
> We should probably extract relevent information from this discussion and add it to the spec text it self: [#37 (comment)](https://github.com/cs3org/OCM-API/pull/37#issuecomment-435875108)

That's true, this comment is even referenced "as is" but it would be good to get more up to date information about current issues with `.well-known`. Agreed that it's for another PR.

And otherwise, with @labkode we better defined how to use DNS `SRV` records, which is a much more viable solution than assuming that an organization can deploy `https://example.com/.well-known/ocm` sort of URLs: the latter are typically not under the control of the EFSS team.

glpatcern:
@DanScharon we slightly modified the logic for discovery: the goal is the same, but typically a https://example.com/.well-known/ocm address is hard to be maintained by the cloud storage team as it "belongs" to the institutional web site and its auth system. Instead, a DNS query can do it, see the new description.

PR #86, opened 2024-04-26
clarify sharedSecret usage

butonic:
We really must not put shared secrets into URLs. They will get logged and might end up who knows where.

This PR clarifies that sharedSecred that the sharedSecret MUST NOT appear in URIs and updates the examples accordingly.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -467,6 +467,7 @@ definitions:
                   description: |
                     An optional secret to be used to access the resource,
                     such as a bearer token.
+                    To prevent leaking it in logs it MUST NOT appear in any URI.
                 permissions:
                   type: array
                   items:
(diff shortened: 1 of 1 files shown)


glpatcern (approved):
This is fully correct, thanks

glpatcern:
Thanks @butonic for spotting this error in the example, the flaw was due at the time to not having specified how to _access_ an OCM share.

Eventually, we have v1.0 doing a `Basic` auth, and v1.1 not specifying anything. IMHO we should enforce a `Bearer` auth using the secret as bearer token (which in the future may be replaced by some other oauth-like token).

And I wonder if we have to amend the published v1.1 specs with all that!

butonic:
@glpatcern ok, so what is the policy on clicking merge ;-)

glpatcern:
Just go ahead and merge ;-) plus please join the call today! I think we should also clarify how access works between 1.0 and 1.1/1.next.

mickenordin (approved):
Very good clarification

PR #90, opened 2024-08-13
/invitation-accepted Reword description

mickenordin:
This patch makes it clearer what the meaning of the status code is


glpatcern (approved):
Looks good as well!

PR #91, opened 2024-08-13
/invitation-accepted remove 404 status

mickenordin:
In #58 this status code was added, I think it ought to be removed in favor of the 400 status code that allready exists, so I made the message clearer.


glpatcern (approved):
Looks good!

PR #92, opened 2024-08-21
Signing request

### LINKED ISSUE #23: Security proposal: trusted services and shared secrets / request signing

ArtificialOwl:
This is an attempt to add an way to confirm the origin of an incoming request when running an OCM request to ensure both party known each others

### Signing Request

Signing request allows to confirm the identity of the sender.
Signing request does not encrypt nor affect its payload.
Signing request only adds metadata to the headers of the request.

### Discovery

A signatory containing a keyId and the public key is added to the OCM discovery

```
{
  "enabled": true,
  "endPoint": "...",
  "resourceTypes": [
    [...]
  ],
  "protocols": {
    [...]
  },
  "publicKey": {
    "keyId": "https://cloud.mydomain.com/ocm#signature",
    "publicKeyPem": "[...]"
  }
}
```

### Signature

The concept is to add unique metadata and sign them using a private/public key pair.
The location of the public key used to verify the signature will confirm the origin of the request.

Signature does not affect the data of the request, it only adds headers to it:

```
 {
     "(request-target)": "post /path",
     "content-length": 380,
     "date": "Mon, 08 Jul 2024 14:16:20 GMT",
     "digest": "SHA-256=U7gNVUQiixe5BRbp4Tg0xCZMTcSWXXUZI2\\/xtHM40S0=",
     "host": "hostname.of.the.recipient",
     "Signature": "keyId=\"https://author.hostname/key\",algorithm=\"ras-sha256\",headers=\"content-length date digest host\",signature=\"DzN12OCS1rsA[...]o0VmxjQooRo6HHabg==\""
 }
```
    'content-length' is the total length of the data/content
    'date' is the datetime the request have been initiated
    'digest' is a checksum of the data/content
    'host' is the hostname of the recipient of the request (remote when signing outgoing request, local on incoming request)
    'Signature' contains the signature generated using the private key, and metadata:
        'keyId' is a unique id, formatted as an url. hostname is used to retrieve the public key via custom discovery
        'algorithm' define the algorithm used to generate signature
        'headers' contains a list of element used during the generation of the signature
        'signature' is the encrypted string, using local private key, of an array containing elements
        listed in 'headers' and their value. Some elements (content-length date digest host) are mandatory
        to ensure authenticity override protection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Since there is no way to guarantee that the sharee OCM provider will actually en
 
 ## Signing request
 
-A request is signed by adding the signature in the headers. The sender also need to expose the public key used to generate the signature. The receiver can then validate the signature and therefore the origin of the request.
+A request is signed by adding the signature in the headers. The sender also needs to expose the public key used to generate the signature. The receiver can then validate the signature and therefore the origin of the request.
 To helps debugging, it is made mandatory to also add all properties used in the signature as headers, even if they can be easily re-generated from the payload.  
 
 Note: Signed requests prove the identity of the sender but does not encrypt nor affect its payload.
--- a/README.md
+++ b/README.md
@@ -181,8 +181,7 @@ Here is an example of how to verify the signature using the headers, the signatu
 
 ### Validating the payload
 
-while signed, it is still needed to confirm the validity of the payload.  
-The last step is to ensure that the payload implies action initiated on the behalf the source of the request.  
+Following the validation of the signature, the host should also confirm the validity of the payload, that is ensuring that the actions implied in the payload actually initiated on behalf of the source of the request.
 
 As an example, if the payload is about initiating a new share the file owner has to be an account from the instance at the origin of the request.  
 
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Here is an example of headers needed to sign a request.
     "date": "Mon, 08 Jul 2024 14:16:20 GMT",
     "digest": "SHA-256=U7gNVUQiixe5BRbp4Tg0xCZMTcSWXXUZI2\\/xtHM40S0=",
     "host": "hostname.of.the.recipient",
-    "Signature": "keyId=\"https://author.hostname/key\",algorithm=\"ras-sha256\",headers=\"content-length date digest host\",signature=\"DzN12OCS1rsA[...]o0VmxjQooRo6HHabg==\""
+    "Signature": "keyId=\"https://author.hostname/key\",algorithm=\"rsa-sha256\",headers=\"content-length date digest host\",signature=\"DzN12OCS1rsA[...]o0VmxjQooRo6HHabg==\""
   }
 ```
 
@@ -141,7 +141,7 @@ This is a quick PHP example of headers for outgoing request:
 		
     $signature = [
         'keyId' => 'https://author.hostname/key',
-        'algorithm' => 'ras-sha256',
+        'algorithm' => 'rsa-sha256',
         'headers' => 'content-length date digest host',
         'signature' => $signed 
     ];
--- a/README.md
+++ b/README.md
@@ -43,8 +43,6 @@ To fill the gap between users knowning other peers' email addresses of the form 
 * A provider SHOULD ensure that a `type=SRV` DNS query to `_ocm._tcp.provider.org` resolves to e.g. `service = 10 10 443 my-cloud-storage.provider.org`
 * When requested to discover the EFSS endpoint for `user@provider.org`, implementations SHOULD query the corresponding `_ocm._tcp.domain` DNS record, e.g. `_ocm._tcp.provider.org`, and subsequently make a HTTP GET request to the host returned by that DNS query, followed by the `/.well-known/ocm` URL path.
 
-
-
 ### Share Creation
 To create a share, the sending server SHOULD make a HTTP POST request to the `/shares` endpoint of the receiving server ([docs](https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org#/paths/~1shares/post)).
 
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ the [OpenAPI](https://github.com/OAI/OpenAPI-Specification) (fka Swagger) specif
   * [Share Updating](#update)
   * [Resharing](#reshare)
   * [Invite](#invite)
+  * [Signing Request](#signing-request)
 
 * [Contributing](#contributing)
 
@@ -35,16 +36,20 @@ If a finite whitelist of receiver servers exists on the sender side, then this l
 
 When a sending server allows sharing to any internet-hosted receiving server, then discovery can happen from the sharee address, using the `/.well-known/ocm` (or `/ocm-provider`, for backwards compatibility) URL that receiving servers SHOULD provide according to this [specification](https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org#/paths/~1.well-known~1ocm/get).
 
+If the identity of an incoming request needs to be confirmed, the discovery data SHOULD contain a public key. Each incoming request that requires to origin from an authenticated source must be signed in its headers using the related private key.
+
 To fill the gap between users knowning other peers' email addresses of the form `user@provider.org`, and the actual cloud storage endpoints being in the form `https://my-cloud-storage.provider.org`, a further discovery mechanism SHOULD be provided by implementations that wish to allow sending shares to any receiver, based on DNS `SRV` Service Records.
 
 * A provider SHOULD ensure that a `type=SRV` DNS query to `_ocm._tcp.provider.org` resolves to e.g. `service = 10 10 443 my-cloud-storage.provider.org`
 * When requested to discover the EFSS endpoint for `user@provider.org`, implementations SHOULD query the corresponding `_ocm._tcp.domain` DNS record, e.g. `_ocm._tcp.provider.org`, and subsequently make a HTTP GET request to the host returned by that DNS query, followed by the `/.well-known/ocm` URL path.
 
+
+
 ### Share Creation
 To create a share, the sending server SHOULD make a HTTP POST request to the `/shares` endpoint of the receiving server ([docs](https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org#/paths/~1shares/post)).
 
 ### Share Acceptance
-In response to a share creation, the receiving server MAY send back a [notification](https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org#/paths/~1notifications/post) to the sending server, with  `notificationType` set to `"SHARE_ACCEPTED"` or `"SHARE_DECLINED"`. The sending server MAY expose this information to the end user. 
+In response to a share creation, the receiving server MAY send back a [notification](https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org#/paths/~1notifications/post) to the sending server, with  `notificationType` set to `"SHARE_ACCEPTED"` or `"SHARE_DECLINED"`. The sending server MAY expose this information to the end user.
 
 ### Share Access
 To access a share, the receiving server MAY use multiple ways, depending on the received payload and on the `protocol.name` property:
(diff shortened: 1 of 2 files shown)


ArtificialOwl:
And an implementation of the feature on Nextcloud: https://github.com/nextcloud/server/pull/45979

glpatcern:
Hi there, interesting proposal indeed, it would nicely complement the recent additions related to discovering the remote end.

Could you please patch the `spec.yaml` file with the required snippet in the discovery endpoint? I'd add the `publicKey` section at the top level of the returned JSON though, not as one of the `resourceTypes`.

ArtificialOwl:
formatted result of the edit in `spec.yaml`

image

artonge (approved):

README.md
```diff
@@ -87,6 +90,41 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+     "Signature": "keyId=\"https://author.hostname/key\",algorithm=\"ras-sha256\",headers=\"content-length date digest host\",signature=\"DzN12OCS1rsA[...]o0VmxjQooRo6HHabg==\""
+ }
+```
+
+- 'content-length' is the total length of the payload of the request
+- 'date' is the datetime the request have been initiated
+- 'digest' is a checksum of the payload of the request
+- 'host' is the hostname of the recipient of the request (remote when signing outgoing request, local on incoming request)
+- 'Signature' contains the signature generated using the private key, and metadata:
+  * 'keyId' is a unique id, formatted as an url. hostname is used to retrieve the public key via custom discovery
```
susnux, line 117: from the specs this important part is missing here:

> [...] The hostname set the origin of the request and **MUST** be identical to the current discovery endpoint

spec.yaml
```diff
@@ -357,6 +357,23 @@ definitions:
...
+            type: string
+            description: |
+              unique id of the key in URI format. The hostname set the origin of the 
+              request and MUST be identical to the current discovery endpoint.
+            example: https://my-cloud-storage.org/ocm#signature
+          publicKeyPem:
+            type: string
+            description: |
+              PEM-encoded version of the public key.
+            example: "-----BEGIN PRIVATE KEY-----\nMII...QDD\n-----END PRIVATE KEY-----\n"
```
```suggestion
            example: "-----BEGIN PUBLIC KEY-----\nMII...QDD\n-----END PUBLIC KEY-----\n"
```

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+- 'content-length' and 'digest' can be re-generated and compared from the payload of the request,
+- A maximum TTL must be applied to 'date' and current timestamp, 
+- regarding data extracted from 'Signature':
+  * using 'keyId' to get the public key from remote signatory,
+  * 'headers' is used to generate the clear version of the signature and must contain at least 'content-length', 'date', 'digest' and 'host',
+  * 'signature' is the encrypted version of the signature.
+
+Using the clear version of the signature, the public key and the encrypted version, it is easy to confirm the authenticity of the request:
+
+```php
```
glpatcern, line 170: I'd suggest to describe this with an abstract pseudo-language, as opposed to PHP and its PHP-specific libraries.

michielbdejong:
Great idea! Side note:
If we want to make OCM compatible with GNAP (see #98), we should require the information advertised via .well-known to match the client details described in section 2.3 of [GNAP](https://datatracker.ietf.org/doc/draft-ietf-gnap-core-protocol/)

michielbdejong:
Also, in GNAP I just read:

> A client instance that is capable of talking to multiple AS's SHOULD
   use a different key for each AS to prevent a class of mix-up attacks
   as described in Section 13.31 unless other mechanisms can be used to
   assure the identity of the AS for a given request.
   
  I wonder if/how that would apply to OCM.

GNAP "client instance" = OCM receiving server
GNAP "Authorization Server (AS)" = OCM sending server

README.md
```diff
@@ -35,16 +36,20 @@ If a finite whitelist of receiver servers exists on the sender side, then this l
...
 
+If the identity of an incoming request needs to be confirmed, the discovery data SHOULD contain a public key. Each incoming request that requires to origin from an authenticated source must be signed in its headers using the related private key.
+
 To fill the gap between users knowning other peers' email addresses of the form `user@provider.org`, and the actual cloud storage endpoints being in the form `https://my-cloud-storage.provider.org`, a further discovery mechanism SHOULD be provided by implementations that wish to allow sending shares to any receiver, based on DNS `SRV` Service Records.
 
 * A provider SHOULD ensure that a `type=SRV` DNS query to `_ocm._tcp.provider.org` resolves to e.g. `service = 10 10 443 my-cloud-storage.provider.org`
 * When requested to discover the EFSS endpoint for `user@provider.org`, implementations SHOULD query the corresponding `_ocm._tcp.domain` DNS record, e.g. `_ocm._tcp.provider.org`, and subsequently make a HTTP GET request to the host returned by that DNS query, followed by the `/.well-known/ocm` URL path.
 
+
+
```
glpatcern:
```suggestion
```

README.md
```diff
@@ -35,16 +36,20 @@ If a finite whitelist of receiver servers exists on the sender side, then this l
 
 When a sending server allows sharing to any internet-hosted receiving server, then discovery can happen from the sharee address, using the `/.well-known/ocm` (or `/ocm-provider`, for backwards compatibility) URL that receiving servers SHOULD provide according to this [specification](https://cs3org.github.io/OCM-API/docs.html?branch=develop&repo=OCM-API&user=cs3org#/paths/~1.well-known~1ocm/get).
 
+If the identity of an incoming request needs to be confirmed, the discovery data SHOULD contain a public key. Each incoming request that requires to origin from an authenticated source must be signed in its headers using the related private key.
```
glpatcern:
```suggestion
To ease the process of confirming the identity of a remote party, the discovery data MAY contain a public key: each incoming request that requires to origin from an authenticated source MUST be signed in its headers using the private key of that source, whose public key MUST be exposed in its discovery data.
```

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
 Since there is no way to guarantee that the sharee OCM provider will actually enforce the MFA requirement, it is up to the sharer OCM provider to establish a trust with the OCM sharee provider such that it is reasonable to assume that the sharee OCM provider will honor the MFA requirement. This establishment of trust will inevitably be implementation dependent, and can be done for example using a pre approved allow list of trusted OCM providers. The procedure of establishing trust is out of scope for this specification: a mechanism similar to the [ScienceMesh](https://sciencemesh.io) integration for the [Invite](#invite) capability may be envisaged.
 
 
+## Signing request
+
+A request is signed by adding the signature in the headers. The sender also need to expose the public key used to generate the signature. The receiver can then validate the signature and therefore the origin of the request.
```
glpatcern:
```suggestion
A request is signed by adding the signature in the headers. The sender also needs to expose the public key used to generate the signature. The receiver can then validate the signature and therefore the origin of the request.
```

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+To helps debugging, it is made mandatory to also add all properties used in the signature as headers, even if they can be easily re-generated from the payload.
```
glpatcern:
```suggestion
To help debugging, it is recommended to also add all properties used in the signature as headers, even if they can easily be re-generated from the payload.  
```

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+    "date": "Mon, 08 Jul 2024 14:16:20 GMT",
+    "digest": "SHA-256=U7gNVUQiixe5BRbp4Tg0xCZMTcSWXXUZI2\\/xtHM40S0=",
+    "host": "hostname.of.the.recipient",
+    "Signature": "keyId=\"https://author.hostname/key\",algorithm=\"ras-sha256\",headers=\"content-length date digest host\",signature=\"DzN12OCS1rsA[...]o0VmxjQooRo6HHabg==\""
+  }
+```
+
+- '(request-target)' contains the reached endpoint and the used method,
+- 'content-length' is the total length of the payload of the request,
+- 'date' is the date and time on which the request have been emitted,
```
glpatcern:
```suggestion
- 'date' is the date and time when the request has been sent,
```

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+- 'digest' is a checksum of the payload of the request,
+- 'host' is the hostname of the recipient of the request (remote when signing outgoing request, local on incoming request),
+- 'Signature' contains the signature generated using the private key and details on its generation:
+  * 'keyId' is a unique id, formatted as an url. hostname is used to retrieve the public key via custom discovery
```
glpatcern, line 122: What does that mean? If the mentioned discovery is querying the OCM `.well-known` endpoint, then this may not be well defined if the hostname is different.

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+    $signed = "DzN12OCS1rsA[...]o0VmxjQooRo6HHabg==";
+    if (openssl_verify(implode("\n", $clear), $signed, $publicKey, 'sha256') !== 1) {
+        throw new InvalidSignatureException('signature issue');
+    }
+```
+
+### Validating the payload
+
+while signed, it is still needed to confirm the validity of the payload.  
+The last step is to ensure that the payload implies action initiated on the behalf the source of the request.
```
glpatcern:
```suggestion
Following the validation of the signature, the host should also confirm the validity of the payload, that is ensuring that the actions implied in the payload actually initiated on behalf of the source of the request.
```

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+        throw new InvalidSignatureException('signature issue');
+    }
+```
+
+### Validating the payload
+
+while signed, it is still needed to confirm the validity of the payload.  
+The last step is to ensure that the payload implies action initiated on the behalf the source of the request.  
+
+As an example, if the payload is about initiating a new share the file owner has to be an account from the instance at the origin of the request.
```
glpatcern, line 189: Apart the minor rephrasing I proposed above, I'm not sure this can actually be done in general, and the example in particular does not seem feasible: how can an EFSS validate something that refers to user belonging to a remote EFSS? Ultimately it has to trust the remote EFSS and its payload.

glpatcern (changes requested):
Hi there, thanks for the extra contribution. I went through the proposal in more details and apart some minor fixes/rephrasing, there are a couple of questions we need to address before proceeding.

MahdiBaghbani:
I've something to add here, it's from the [ActivityPub](https://www.w3.org/wiki/ActivityPub/Primer/Authentication_Authorization),


> Signing requests using HTTP Signatures
Server to server federation via [inbox delivery POST requests](https://www.w3.org/TR/activitypub/#delivery) is authenticated using [HTTP Signatures](https://tools.ietf.org/html/draft-cavage-http-signatures-08) in conjunction with the signing key from the actor's [publicKey](https://web-payments.org/vocabs/security#publicKey) field. The [keyId](https://tools.ietf.org/html/draft-cavage-http-signatures-08#section-2.1.1) should link to the actor so that the publicKey field can be retrieved. At minimum, the digest field should be included in the set of headers being signed.

So the actors are the users on the server, and each user has a key pair. The requests are signed with the private key of the user initiating the request (and not the server).

There is also an `Instance Actor` which is the server itself.

README.md
michielbdejong: https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12#section-2.1.1
Maybe we should say in our case keyId should be an FQDN? And then we use .well-known with fallback to DNS SRV from that FQDN.

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+Here is an example of headers needed to sign a request.  
+
+```
+  {
+    "(request-target)": "post /path",
+    "content-length": 380,
+    "date": "Mon, 08 Jul 2024 14:16:20 GMT",
+    "digest": "SHA-256=U7gNVUQiixe5BRbp4Tg0xCZMTcSWXXUZI2\\/xtHM40S0=",
+    "host": "hostname.of.the.recipient",
+    "Signature": "keyId=\"https://author.hostname/key\",algorithm=\"ras-sha256\",headers=\"content-length date digest host\",signature=\"DzN12OCS1rsA[...]o0VmxjQooRo6HHabg==\""
```
michielbdejong: Shouldn't headers be `"(request-target) content-length date digest host"`? (see also [line 135](https://github.com/cs3org/OCM-API/pull/92/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R135)).
And you have a typo I think, `algorithm` should be `"rsa-sha256"`

README.md
```diff
@@ -87,6 +92,103 @@ If an OCM provider exposes the capability `/mfa-capable`, it indicates that it w
...
+        'date' => gmdate('D, d M Y H:i:s T'),
+        'digest': 'SHA-256=' . base64_encode(hash('sha256', utf8_encode($payload), true)),
+        'host': 'hostname.of.the.recipient',
+    ];
+
+    openssl_sign(implode("\n", $headers), $signed, $privateKey, OPENSSL_ALGO_SHA256);
+		
+    $signature = [
+        'keyId' => 'https://author.hostname/key',
+        'algorithm' => 'ras-sha256',
```
michielbdejong: rsa-sha256

michielbdejong:
We should merge this into the spec as optional, great stuff!
I implemented it in OcmStub for both sending and receiving.

glpatcern:
I concur this is a most welcome development to be merged to the spec.

I also linked https://github.com/cs3org/OCM-API/issues/23 from a few years back as it suggests exactly the same thing.

michielbdejong:
An [RFC](https://datatracker.ietf.org/doc/html/rfc9421) came out in February, looks like this is intended to replace [the I-D from 2019](https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12)?

README.md
michielbdejong: Let's stick to URL as proposed here, we can discuss in a follow-up whether FQDN is more appropriate for us.

README.md
michielbdejong: Sure, but I think what is meant is: if the owner is specified as `alice@cernbox.cern.ch` but the sender is `cloud.pondersource.com` then the receiving server should flag the domain name mismatch.

We could also make this automatic in a future version of the spec, so that owner would just be `alice` and the `@cernbox.cern.ch` would be implied silently from the keyId of the signature.

README.md
michielbdejong: These are details about [httpsig](https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12), we can fix this in a follow-up.
I also noticed that node refuses to add `(request-target)` as a header (saying it is malformed), so in my implementation I changed it to `request-target`.

michielbdejong (approved):

spec.yaml
```diff
@@ -357,6 +357,23 @@ definitions:
           enum: ["/notifications", "/invite-accepted", "/mfa-capable"]
         example:
           ["/invite-accepted"]
+      publicKey:
+        type: object
+        description: |
+          The signatory used to sign outgoing request to confirm its origin. The 
+          signatory is optional but it MUST contains `id` and `publicKeyPem`.
```
```suggestion
          signatory is optional but it MUST contain `id` and `publicKeyPem`.
```

spec.yaml
```diff
@@ -357,6 +357,23 @@ definitions:
...
+        properties:
+          id:
+            type: string
+            description: |
+              unique id of the key in URI format. The hostname set the origin of the
```
glpatcern, line 369: Is this intended to read `The hostname is to be set to the origin` ? In that case, is that redundant? or is it still better to have it for "standardization" purposes?

mickenordin (approved):
This looks good to me now.

glpatcern (approved):
Only a few minor fixes remaining.

Otherwise, I don't think it's good to have language-specific code examples in the spec. If you want to move forward let's merge this as is, but please open an issue that the example on how to perform the signing and the verification should be done in a more abstract way, especially without language-specific primitives such as `implode()` (Let's say up to `base64_encode(hash('sha256', utf8_encode($payload)))` is fine!).

spec.yaml
michielbdejong: I think we should actually use FQDN here.

michielbdejong: I'll merge this now as it is proposed by @ArtificialOwl and then create my own follow-up PR to iterate on this.

michielbdejong:
> If you want to move forward let's merge this as is

Thanks! Yes, let's merge this now and then I'll add my own ideas in a follow-up PR

spec.yaml
glpatcern: ok fine, so also the phrasing can be made clearer

ArtificialOwl:
sorry, was overhelmed with work in the last few weeks, not even able to free me for those weekly calls

PR #93, opened 2024-08-22
Advertise any other protocol in the discovery endpoint

glpatcern:
This is a proposal in order to standardize how to show additional protocols in the discovery endpoint.

~~I'm not sure it can be espressed in OpenAPI 2.0. One more reason to convert to OpenAPI 3.0?~~

Edit: I fixed the PR (still with OpenAPI 2.0) so to allow to specify `<any name> : "<url>"`, similar to how Nextcloud advertises their `talk` protocol:
```
      "protocols": {
        "talk-v1": "/ocs/v2.php/apps/spreed/api/"
      }
```
And the included example is deliberately inspired to the above.

@ArtificialOwl this makes Nextcloud's discovery fully standard ;-)

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -320,8 +320,9 @@ definitions:
             protocols:
               type: object
               description: |
-                The supported protocols to access shares at this endpoint.
-                Implementations MUST support `webdav` at a minimum.
+                The supported protocols to access shared resources at this endpoint.
+                Implementations MUST support at least `webdav` for `file` resources,
+                any other combination of resources and protocols is optional.
               properties:
                 webdav:
                   type: string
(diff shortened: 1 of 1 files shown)


MahdiBaghbani:
> But I'm not sure it can be espressed in OpenAPI 2.0. One more reason to convert to OpenAPI 3.0?

It is possible to do that.

glpatcern:
For reference, this is the discovery endpoint of Nextcloud:

https://cloud.nextcloud.com/ocm-provider

MahdiBaghbani:
https://github.com/cs3org/OCM-API/blob/3b419333998439ce4b58893f60c5c44083735026/spec.yaml#L438-L546

michielbdejong (approved):

PR #95, opened 2024-08-22
Federated contacts public key exchange and signing

MahdiBaghbani:
In addition to #92 I'd like to create a PR for a similar matter.

I propose to:
1. Sign the requests on behalf of the sender instance (which seems to be **required**).
2. "**Optionally**" also sign the requests on behalf of the sender user and receiver user.

This could be beneficial in:
1. Make sure the sender is the user it claims to be (even if the sender server is authentic)
2. It allows sharing E2EE shares from one vendor to another (this one needs discussion).

Cons:
1. This only works if users do the invite-flow first and then try to share something.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -640,6 +640,27 @@ definitions:
             type: string
             description: Name of the user accepting the invite.
             example: Richard Feynman
+          publicKey:
+            type: object
+            description: Public key of the user accepting the invite.
+            properties:
+              keyID:
+                type: string
+                format: url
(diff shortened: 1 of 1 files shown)


michielbdejong:
Hm, good idea but feels a bit out of scope, and feels like it should be a mechanism that works both for OCM and for Toots and other social notifications and messages. Can we propose this as a separate spec at the fediverse level? Maybe a separate NLnet project even?

MahdiBaghbani:
If you have time we can discuss this as a separate NLnet project in this week.

PR #96, opened 2024-08-26
Make DNS-based discovery optional

michielbdejong:
In the current text, it looked like DNS-based discovery should always happen before the .well-known query. But that would break all existing implementations. This PR changes it so that it is only a fallback mechanism.


glpatcern (approved):
Nice! This was the intention in the first place, and the new text makes it clearer.

PR #98, opened 2024-08-26
httpsig and short-lived bearer tokens as alternative to sharedSecret

michielbdejong:
This would allow for some more modern security best practices like pre-registering clients and making the access token short-lived and client-bound

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,9 @@ In response to a share creation, the receiving server MAY send back a [notificat
 To access a share, the receiving server MAY use multiple ways, depending on the received payload and on the `protocol.name` property:
 
 * If `protocol.name` = `multi`, the receiver MUST make a HTTP PROPFIND request to `protocol.webdav.uri` to access the remote share. If `protocol.webdav.sharedSecret` is not empty, the receiver MUST pass it as a `Authorization: bearer` header.
-Otherwise, if `protocol.webdav.code` and `protocol.webdav.token_endpoint` are not empty, the receiver SHOULD call the token endpoint to exchange
-the code for a short-lived access token, using an [Access Token Request](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3)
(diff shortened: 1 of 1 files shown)


michielbdejong:
I think we should use a GNAP grant request instead. Will update.

glpatcern (approved):
Makes sense. Looking forward to see how this section about the _access procedure_ can be made more "standardized" compared to the current free-form text!

michielbdejong:
We can reuse the format of the WWW-Authenticate header defined for GNAP in section 9.1 of https://datatracker.ietf.org/doc/draft-ietf-gnap-core-protocol/

so then it would be `as_uri` instead of `gnapAuthorizationServer`. I'll work on this some more.

michielbdejong:
See https://github.com/cs3org/ocm-test-suite/issues/88#issuecomment-2325970815 for a demo of how this would work

README.md
```diff
@@ -50,6 +50,9 @@ In response to a share creation, the receiving server MAY send back a [notificat
 To access a share, the receiving server MAY use multiple ways, depending on the received payload and on the `protocol.name` property:
 
 * If `protocol.name` = `multi`, the receiver MUST make a HTTP PROPFIND request to `protocol.webdav.uri` to access the remote share. If `protocol.webdav.sharedSecret` is not empty, the receiver MUST pass it as a `Authorization: bearer` header.
+Otherwise, if `protocol.webdav.code` is not empty, the receiver SHOULD discover the sender's `tokenEndpoint` and make a signed POST request to it, to exchange
```
glpatcern: How is `tokenEndpoint` defined in this context?

michielbdejong: I just changed that a bit just now - if the endpoint is `/ocm` then the token endpoint will be `/ocm/token`.

glpatcern:
Nice to see you already have your ocm-stub doing this. Would you add `protocol.webdav.code` in the `spec.yaml` as another optional parameter?

PR #101, opened 2024-08-28
Add sharedSecret back in multipleProtocols example

michielbdejong:


glpatcern (approved):
Yes, this was overlooked in #86, this PR ought to be merged to keep the example consistent with the spec.

PR #102, opened 2024-08-31
Use federatedCloudId instead of email for user identification between servers

MahdiBaghbani:
This PR changes 2 things:

1. Renames `userID` to `userId` to be consistent with `camelCase` across the whole specification eg. `providerId`
2. Adds a field to the `invite-flow` called `federatedCloudId`.
3. Removes `recipientProvider` since it is redundant and is repeated on `federatedCloudId`.

---

**Reason behind the new field**
Previously OCM used email as a way to identify remote users. 

This is essential when a party tries to send a `share` or `notification` to another party because it is going to find the domain name of the recipient from their ID (which currently is set to email, look at `Reva` and `ScienceMesh`)

This becomes problematic when the user's email name or domain is different than the user's username or provider domain. for example:

1. Alice is on provider `apollo.space` with username `alice1998` and has email `alice1998@apollo.space` -> OK
2. Bob is on provider `nasa.space` with username `bob` and has email `bobby@gmail.com` -> Problem

Vendors can overcome this problem by not providing the user's email and just returning username@provider-domain in the email field.

I think it is a good idea to make a distinction between email and `federatedCloudId`.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -654,7 +654,7 @@ definitions:
             example: 9302
           federatedCloudId:
             type: string
-            description: Federated cloud ID of the user accepting the invite. it should be in the form of username@provider-domain.
+            description: Federated cloud ID of the user who sent the invite. it should be in the form of username@provider-domain.
             example: john@sender.org
           email:
             type: string
--- a/spec.yaml
+++ b/spec.yaml
@@ -619,11 +619,6 @@ definitions:
     type: object
     allOf:
       - properties:
-          recipientProvider:
-            type: string
-            format: url
-            description: URL of the receiver OCM service.
-            example: https://receiver.org
           token:
             type: string
             description: Token received in the invite
--- a/spec.yaml
+++ b/spec.yaml
@@ -628,14 +628,18 @@ definitions:
             type: string
             description: Token received in the invite
             example: xyz
-          userID:
+          userId:
             type: string
             description: Unique ID to identify the user at the remote provider accepting the invite.
             example: 51dc30ddc473d43a6011e9ebba6ca770
+          federatedCloudId:
+            type: string
+            description: Federated cloud ID of the user accepting the invite. it should be in the form of username@provider-domain.
+            example: richard@receiver.org
           email:
             type: string
-            description: Email ID of the user accepting the invite.
-            example: richard@receiver.org
+            description: Email of the user accepting the invite.
+            example: richard@gmail.com
           name:
             type: string
             description: Name of the user accepting the invite.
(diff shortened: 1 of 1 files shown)


glpatcern (approved):
This is an interesting proposal as the net sum is to add the `username` of a remote user as known by the remote EFSS, on top of the `email` identifier that indeed may not be related to the remote EFSS' domain.

I think some comment from @labkode as originally involved in the conception, as well as @michielbdejong, is more than welcome prior to merging this enhancement.

spec.yaml
```diff
@@ -619,23 +619,22 @@ definitions:
...
-          userID:
+          userId:
             type: string
             description: Unique ID to identify the user at the remote provider accepting the invite.
             example: 51dc30ddc473d43a6011e9ebba6ca770
-          email:
+          federatedCloudId:
             type: string
-            description: Email ID of the user accepting the invite.
+            description: Federated cloud ID of the user accepting the invite. it should be in the form of username@provider-domain.
```
michielbdejong, line 632: I'm confused by this, what is the difference between username `username` and the userId `51dc30ddc473d43a6011e9ebba6ca770` in this context?

Is this invite being accepted by `richard@receiver.org` or by `51dc30ddc473d43a6011e9ebba6ca770@receiver.org`?

michielbdejong (changes requested):
Let's not merge this until we understand the relationship between the proposed `federatedCloudId` and the existing `userId`

MahdiBaghbani:
It seems, that the invite is being accepted by `51dc30ddc473d43a6011e9ebba6ca770@receiver.org`

As implemented in [ScienceMesh](https://github.com/sciencemesh/nc-sciencemesh/blob/66b8562a4676f22a8ef102eba732c1852f3a6065/lib/RevaHttpClient.php#L108-L142)

So no need for the new filed in this PR.

spec.yaml
MahdiBaghbani: It was accepted by `51dc30ddc473d43a6011e9ebba6ca770@receiver.org`.

MahdiBaghbani:
I think this `https://` (look at [Test Suite](https://github.com/pondersource/dev-stock/issues/128)) bug is originated from here.

The `"recipientProvider": "https://receiver.org/"` has the protocol schema instead of just the domain.

Proposal: 
Merge both `userId` and `recipientProvider` into `cloudId` = `userId@recipientProviderDomain`

michielbdejong:
> Proposal:
Merge both `userId` and `recipientProvider` into `cloudId` = `userId@recipientProviderDomain`

That would be a breaking spec change with costs to implementers. What if instead we just improve how the existing `userId` and `recipientProvider` fields are documented? In particular, make it clear that `recipientProvider` should be an FQDN and not a URL?

MahdiBaghbani:
That's right, that makes sense since it is already inside Reva and ScienceMesh, and they don't check if it is URL or FQDN.

PS. For future reference, this is the [Reva](https://github.com/cs3org/reva/blob/8d1b4abe872c2fc0a8f1734f12d2b09512c93c05/internal/http/services/opencloudmesh/ocmd/invites.go#L72-L75) code for accepting an invite.

michielbdejong:
So do we want to close this PR without merging then?

MahdiBaghbani:
Yes!

> we just improve how the existing userId and recipientProvider fields are documented.

that would be in another PR.

PR #108, opened 2024-09-04
Remove required but not existing properties from NewShare?

michielbdejong:
It seems these required properties don't exist? But it seems https://editor.swagger.io/ doesn't detect this as an error, am I seeing this wrong?

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -374,6 +374,8 @@ definitions:
       - providerId
       - owner
       - sender
+      - shareType
+      - resourceType
       - protocol
     properties:
       shareWith:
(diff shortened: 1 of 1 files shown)


MahdiBaghbani:
I can only see that `- permission` doesn't exist in the NewShare.

michielbdejong:
Ah, good catch! So the list is (with **bold** for required fields):
* **shareWith**
* **name**
* description
* **providerId**
* **owner**
* **sender**
* ownerDisplayName
* senderDisplayName
* **shareType**
* **resourceType**
* expiration
* **protocol**

MahdiBaghbani (approved):

michielbdejong:
Also, thinking functionally, adding a permission field in an invite-accepted message doesn't really make much sense. I think it was just there by mistake, especially since it only appeared in the 'required' list and not in the 'properties' list

PR #110, opened 2024-09-04
recipientProvider should be an FQDN

michielbdejong:
Follow-up from #102
Also pick an off-domain email address to highlight that possibility


mickenordin (approved):
LGTM

MahdiBaghbani (approved):

PR #114, opened 2024-09-04
Slight section rewrite: Accessing the Resource

michielbdejong:
I put `code` in front of `protocol.webdav.token` so if both are present, then `code` should be preferred.

I also marked the MFA and Request Signing sections as Appendices of the Specification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,19 +127,39 @@ Following these step, both servers MAY display the `name` of the other party as 
 
 For further details on this concept, see also [#54](https://github.com/cs3org/OCM-API/pull/54) and related issues. For a discussion about trust policies, see [sciencemesh#196](https://github.com/sciencemesh/sciencemesh/issues/196).
 
-### Discovery
-Authentication between services is already established. This means that this specification doesn't cover the way a service authenticates incoming API calls (e.g. through an API Key, VPN connection or IP whitelisting). In this scope we assume that the services are already authenticated.
+### OCM API Discovery
+After establishing contact as discussed in the previous section, the Sharing User can send the Share Creation Gesture to the Sending Server, providing the Sending Server with the following information:
+* Resource to be shared
+* Protocol to be offered for access
+* Sending Party's identifier
+* Receiving Party's identifier
+* Receiving Server FQDN
+* OPTIONAL: Name
+* OPTIONAL: Permissions
 
-If a finite whitelist of receiver servers exists on the sender side, then this list may already contain all necessary endpoint details.
+The next step is for the Sending Server to additionally discover:
+* if the Receiving Server is trusted
+* if the Receiving Server supports OCM
+* if so, which version and with which optional functionality
+* at which URL
+* the public key the Receiving Server will use for HTTP Signatures (if any)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -216,13 +217,13 @@ receiving server to persuarde the sending server to share the same resource with
 TODO: document how receiver.com can know if sender.com understood and processed the
 reshare request.
 
-### Multi Factor Authentication
+## Appendix A: Multi Factor Authentication
 If an OCM provider exposes the capability `/mfa-capable`, it indicates that it will try and comply with a MFA requirement set as a permission on a share. If the sharer OCM provider trusts the receiver OCM provider, the sharer MAY set the permission `mfa-enforced` on a share, which SHOULD be honored. A compliant OCM provider that signals that it is MFA-capable MUST not allow access to a resource protected with the `mfa-enforced` permission, if the consumer has not provided a second factor to establish their identity with greater confidence.
```
mickenordin: Should we switch to `sending server` and `receiving server` in this section, to conform with new definitions?

mickenordin (commented):
Left a comment about sending/receiving server we might want to change

MahdiBaghbani (approved):

README.md
michielbdejong: Yes! Done.

README.md
```diff
@@ -188,17 +188,18 @@ They could give the Receiving Party the option to accept or reject the share, or
...
 
-### Share Access
-To access a share, the receiving server MAY use multiple ways, depending on the received payload and on the `protocol.name` property:
+### Resource Access
+To access the Resource, the Receiving Server MAY use multiple ways, depending on the body of the Share Creation Notification and on the `protocol.name` property in there:
 
-* If `protocol.name` = `multi`, the receiver MUST make a HTTP PROPFIND request to `protocol.webdav.uri` to access the remote share. If `protocol.webdav.sharedSecret` is not empty, the receiver MUST pass it as a `Authorization: bearer` header.
-Otherwise, if `protocol.webdav.code` is not empty, the receiver SHOULD discover the sender's OCM endpoint and make a signed POST request to `<OCM endpoint>/token`, to exchange
+* If `protocol.name` = `multi`, the receiver MUST make a HTTP PROPFIND request to `protocol.webdav.uri` to access the remote share.
+If `code` is not empty, the receiver SHOULD discover the sender's OCM endpoint and make a signed POST request to the `/token` path inside the Sending Server's OCM API, to exchange
```
glpatcern, line 228:
```suggestion
If `protocol.webdav.code` is not empty, the receiver SHOULD discover the sender's OCM endpoint and make a signed POST request to the `/token` path inside the Sending Server's OCM API, to exchange
```

glpatcern (changes requested):
Looks good to me, but the `code` property is not yet defined in the spec?

README.md
michielbdejong: ah yes, well actually in my draft implementation of it in Ocm Stub I figures that the token endpoint is not specific to the resource access protocol - if a Sending Server offers various protocols then it would still probably have only one code per share and only one token endpoint at which this can be exchanged for an access token, to be used in a protocol-specific way.

So that's why I moved it to the base level. We can still move it back into webdav and other protocols in a follow-up PR if we see fit, as long as we take a final decision about it before tagging version 1.2

michielbdejong:
> Looks good to me, but the code property is not yet defined in the spec?

Good point! Fixed in [643fdbd](https://github.com/cs3org/OCM-API/pull/114/commits/643fdbd612983b09e8bb61ab8f1aa3124077fd08)

See https://cs3org.github.io/OCM-API/docs.html?branch=accessing-the-resource&repo=OCM-API&user=cs3org#/paths/~1token/post

glpatcern (approved):
This is good enough for now and further refinements can be made as we keep reviewing the whole text

PR #123, opened 2024-09-05
Change response type

### LINKED ISSUE #120: switch content-type from application/hal+json to application/json
It seems the application/hal+json mention comes from https://github.com/cs3org/OCM-API/issues/12
I don't think we need it (anymore)

Let's remove it and just use application/json instead?

MahdiBaghbani:
fixes https://github.com/cs3org/OCM-API/issues/120

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -10,7 +10,7 @@ schemes:
 consumes:
   - application/json
 produces:
-  - application/hal+json
+  - application/json
 parameters:
   id:
     name: id


michielbdejong (approved):
Yes,  I think the `hal+` brings us nothing and it only confuses people, so it's indeed better to remove it. As discussed in yesterday's weekly meeting

glpatcern (approved):

Issue #104, opened 2024-09-03
Remove `protocol.name`


michielbdejong:
It occured to me that except for the `singleProtocolLegacy` in https://cs3org.github.io/OCM-API/docs.html?branch=v1.1.0&repo=OCM-API&user=cs3org#/paths/~1shares/post we don't really need the `"name": "webdav"` or `"name": "multi"`, right? Can we remove it in `singleProtocolNew` and `multipleProtocols`?

glpatcern:
I definitely remember a lot of resistance on that, because removing `name` essentially is a breaking change w.r.t. OCM 1.0.

glpatcern:
Following my comment in https://github.com/cs3org/OCM-API/issues/127, I think we can deprecate `protocol.name` in the prospect 1.2, and remove it with OCM 2.0.

PR #131, opened 2024-09-12
Reworded the protocols section of a new share

### LINKED ISSUE #125: Can we drop `singleProtocolNew`?
Related to #104.

Looking at [the API spec](https://github.com/cs3org/OCM-API/blob/ddb16c55ff810e910be1ebd826f9731020538a08/spec.yaml#L572-L592) there are 3 ways to specify that WebDAV and nothing else is offered:

1. name: 'webdav', details in 'options' (`singleProtocolLegacy`)
2. name: 'webdav', details in 'webdav' (`singleProtocolNew`) -> why do we need this?
3. name: 'multi', details in 'webdav' (`multipleProtocols`)

Can we remove the second one?
Note that if we decide to drop `name` in #104 then 2. and 3. automatically become identical.

glpatcern:
As discussed I reworded the `protocols` section, and removed some inconsistencies.

This fixes #125 and #104 (the removal of `protocol.name` is postponed to OCM 2.0).

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -494,13 +494,13 @@ definitions:
       expiration:
         type: integer
         description: |
-          The expiration time for the OCM share, in seconds
-          of UTC time since Unix epoch. If omitted, it is assumed
-          that the share does not expire.
+          The expiration time for the share, in seconds of UTC time since
+          Unix epoch. If omitted, it is assumed that the share does not expire.
       code:
         type: string
         description: |
-          A nonce to be exchanged for a (potentially short-lived) bearer token at the Sending Server's /token endpoint.
+          A nonce to be exchanged for a (potentially short-lived) bearer token
+          at the Sending Server's `/token` endpoint.
       protocol:
         type: object
         description: |
@@ -529,6 +529,7 @@ definitions:
                 Warning: client implementers should be aware that v1.1 servers
                 MAY support both `webdav` and `multi`, but v1.0 servers MAY
                 only support `webdav`.
+                This field may be removed in a future major version of the spec.
             options:
               type: object
               description: |
@@ -561,10 +562,8 @@ definitions:
                   type: string
                   description: |
                     An URI to access the remote resource. The URI MAY be relative,
-                    in which case the prefix exposed by the `/ocm-provider` endpoint MUST
-                    be used, or it may be absolute (recommended). Additionally, the URI
-                    MAY include a secret hash in the path, in which case there MAY be
-                    no associated `sharedSecret`.
+                    in which case the prefix exposed by the `/.well-known/ocm` endpoint MUST
+                    be used, or it may be absolute (recommended).
             webapp:
               type: object
               properties:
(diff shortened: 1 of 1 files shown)


spec.yaml
```diff
@@ -604,29 +608,32 @@ definitions:
...
           singleProtocolLegacy:
             name: "webdav"
             options:
               sharedSecret: "hfiuhworzwnur98d3wjiwhr"
               permissions: "some permissions scheme"
           singleProtocolNew:
-            name: "webdav"
-            options:
-              sharedSecret: "hfiuhworzwnur98d3wjiwhr"
+            name: "multi"
```
mickenordin, line 636: Is this correct? Should maybe be `single`?

spec.yaml
```diff
@@ -529,6 +529,7 @@ definitions:
                 Warning: client implementers should be aware that v1.1 servers
                 MAY support both `webdav` and `multi`, but v1.0 servers MAY
                 only support `webdav`.
+                This field may be removed in a future major version of the spec.
```
michielbdejong, line 532:
```suggestion
                Default: `multi`. This field may be removed in a future major version of the spec.
```

michielbdejong: It's clear to us but maybe not to the reader, that if this field is omitted, then `multi` is assumed.

spec.yaml
```diff
@@ -590,8 +589,13 @@ definitions:
                 sharedSecret:
                   type: string
                   description: |
-                    An optional secret to be used to access the remote web app,
-                    for example in the form of a bearer token.
+                    An optional secret to be used to access the remote web app, such as
+                    a bearer token. To prevent leaking it in logs it MUST NOT appear
+                    in any URI. If a `code` is provided, then the sending host MUST
+                    accept the short-lived bearer token when serving the web app,
+                    which was already exchanged on the first WebDAV-based interaction
```
michielbdejong: Why is this first WebDAV-based interaction mandatory?
```suggestion
                    which can be exchanged in the code flow interaction
```

spec.yaml
michielbdejong: You're right that `single` would be more accurate, but putting that value here for this case would just force implementers to add extra if statements, and not convey any information that cannot already be derived by simply counting the keys in the object. If we just always set `multi` as the name by default, then it becomes easy to ignore the name field and just look at the actual entries that are present.

michielbdejong: So I think `multi` is correct here.

michielbdejong (approved):

spec.yaml
glpatcern: It is not mandatory, but it is likely what would happen in a real implementation (granted that the phrasing should change). I should probably create a Sequence Diagram for this, but the idea is the Receiver first accesses the remote share (via WebDAV), and then clicks on some "Open in Remote app" for opening a given resource that is part of that share.

I see the two gestures as sequential, from a UI perspective the only way to jump to "Open in Remote app" would be if the receiver caches (e.g. in a favorite?) the reference of the remote resource.

michielbdejong: You see them as sequential, because of a dependency? I don't think I understand this.

The 'open in remote app' is especially useful if the file is in an app-specific format that the receiving server cannot display (for instance a Jupyter notebook or an excel sheet or a binary measurements file, etc). So fetching the file over WebDAV makes no sense then, it would not be able to display the file anyway.

Why couldn't you go straight from 'you received a share' to 'open in remote app' and skip the fetching?

glpatcern: The dependency - and sequence - I see is purely from a UI perspective, you're right that you could go straight to "open in remote app" from the received share.

What I mean is the user would need to be presented with a list of files in the received share (which requires a WebDAV access), prior to be able to trigger the "open in remote app" action on one of those files. And even in the case of a single-file share, the user will have to have a view of the shared resource as obtained following a `PROPFIND`, and then can the action be triggered.

PR #132, opened 2024-09-13
MFA: Change from permission to a new concept `requirement`

mickenordin:
A requirement is a restriction on a resource, an anti permission if you
will. Currently two types exist, `mfa-enforced` and `none`.
    
Requirements are required to be set on a resource, so that an
implementor can check requirements without having to check if the
attribute is set in the request beforehand.

Please note that the token endpoint removed in the first commit is a duplicate, so it is still in the spec :)


spec.yaml
```diff
@@ -553,10 +534,21 @@ definitions:
...
                       - `write` allows create, update, and delete rights on the resource.
                       - `share` allows re-share rights on the resource.
+                    enum: ["read", "write", "share"]
+                requirements:
+                  type: array
+                  items:
+                    type: string
+                    description: |
+                      A list of requirements that the recipient provider MUST fulfill
+                      to access the resource. The share MUST contain at least the `none`
```
glpatcern: If `requirements` is optional, shall we instead consider that `none` is implicit? And yet, if a `requirements` entry is there, it MUST NOT be empty?

README.md
```diff
@@ -379,7 +381,7 @@ TODO: document how the Receiving Party can know if the Sending Party understood
 reshare request.
 
 ## Appendix A: Multi Factor Authentication
-If a Receiving Server exposes the capability `/mfa-capable`, it indicates that it will try and comply with a MFA requirement set as a permission on a Share. If the Sending Server trusts the Receiving Server, the Sending Server MAY set the permission `mfa-enforced` on a Share, which the Receiving Server SHOULD honor. A compliant Receiving Server that signals that it is MFA-capable MUST not allow access to a resource protected with the `mfa-enforced` permission, if the Receiving Party has not provided a second factor to establish their identity with greater confidence.
+If a Receiving Server exposes the capability `/mfa-capable`, it indicates that it will try and comply with a MFA requirement set on a Share. If the Sending Server trusts the Receiving Server, the Sending Server MAY set the permission `mfa-enforced` on a Share, which the Receiving Server SHOULD honor. A compliant Receiving Server that signals that it is MFA-capable MUST not allow access to a resource protected with the `mfa-enforced` requirement, if the Receiving Party has not provided a second factor to establish their identity with greater confidence.
```
glpatcern: This is OK for now, eventually the capability is going to be changed to not have the form of an endpoint, right?

glpatcern (commented):
Nice, just a couple of comments, as well as a more general question: I'd introduce a `signed` requirement as well, as mentioned yesterday. After all it does not cost much - requirements are optional - but given that signatures are ALSO optional for backwards compatibility, there's a way to enforce them if needed.

README.md
```diff
@@ -92,7 +92,8 @@ unique at the server. `<fqdn>` is the Fully Qualified Domain Name by which the s
 * __Invite Acceptance Request__ - API call from the Invite Receiver OCM Server to the Invite Sender OCM Server, supplying the Invite Token as well as the OCM Address of the Invite Receiver, effectively allowlisting the Invite Sender OCM Server for sending Share Creation Notifications to the Invite Receiver OCM Server.
 * __Invite Acceptance Response__ - HTTP response to the Invite Acceptance Request
 * __Share Name__ - a human-readable string, provided by the Sending Party or the Sending Server, to help the Receiving Party understand which Resource the Share grants access to
-* __Share Permissions__ - protocol-specific restrictions on the modes of accessing the Resource
+* __Share Permissions__ - protocol-specific freedoms granted to the Receiving Party on the modes of accessing the Resource
```
michielbdejong:
```suggestion
* __Share Permissions__ - protocol-specific allowances granted to the Receiving Party on the modes of accessing the Resource
```

README.md
```diff
@@ -361,7 +363,7 @@ Otherwise, if `protocol.webdav.sharedSecret` is not empty, the receiver MUST pas
 
 In both cases, when the Resource is a folder and the Receiving Server accesses a resource within that shared folder, it SHOULD append its relative path to that URL.
 
-Additionally, if `protocol.<protocolname>.permissions` include `mfa-enforced`, the Receiving Server MUST ensure that the Receiving Party has been authenticated with MFA.
+Additionally, if `protocol.<protocolname>.webdav.requirements` include `mfa-enforced`, the Receiving Server MUST ensure that the Receiving Party has been authenticated with MFA.
```
michielbdejong: If the variable `<protocolname>` has the value `'webdav'` then this would become `protocol.webdav.webdav.requirements`?

README.md
michielbdejong: Yes, let's keep track of that, I created #133 about it.

spec.yaml
```diff
@@ -252,27 +254,6 @@ paths:
...
-            $ref: "#/definitions/TokenRequest"
-      responses:
-        200:
-          description: Token issued.
-          schema:
-            $ref: "#/definitions/TokenResponse"
-        403:
-          description: Token denied.
-          schema:
-            $ref: "#/definitions/Error"
```
michielbdejong, line 275: Merge conflict? This is the new token endpoint, used in the code flow, so this PR shouldn't remove it.

spec.yaml
michielbdejong: And maybe add that (while it's OK to ignore the permissions of a share) the Receiving Server MUST reject a share whose requirements it doesn't understand.

michielbdejong (approved):
LGTM

spec.yaml
michielbdejong: Ah right, you're actually fixing a merge conflict, the section was in there twice in the `develop` branch :)

michielbdejong:
@mickenordin can you have a look at the comments and merge this?

README.md
mickenordin: I think this was a logical error on my part, and a misunderstanding on how the example in the spec about how single protocol legacy/ multi protocol was worded. Should be fixed now.

mickenordin:
> Nice, just a couple of comments, as well as a more general question: I'd introduce a `signed` requirement as well, as mentioned yesterday. After all it does not cost much - requirements are optional - but given that signatures are ALSO optional for backwards compatibility, there's a way to enforce them if needed.

I would prefer if we add the `signed` requirement in a separate pr, it is not clear to me if a requirement must be accompanied by a capability. We can discuss that in the next meeting.

glpatcern:
> I would prefer if we add the `signed` requirement in a separate pr, it is not clear to me if a requirement must be accompanied by a capability. We can discuss that in the next meeting.

All right, and yes I believe it must - as for the `mfa`!

PR #136, opened 2024-09-16
Caps Disco

### LINKED ISSUE #121: Discovery caps - Aka is the /mfa-capable endpoint really necessary?
If '/mfa-capable' is listed as a capability in the Discovery Document, then that already states that MFA requirements are supported, right? So then why would a separate boolean endpoint at `/mfa-capable` be necessary?

### LINKED ISSUE #127: Did we introduce protocol.[webdav|options].URI accidentally
I was thinking about the URI property in WebDAV protocol details, and it occurred to me that
1) current implementations ignore it, so for them it's a breaking change
2) I don't think we actively took a decision to introduce this as a breaking change

I traced it back to https://github.com/cs3org/OCM-API/commit/c54058bcd6e6b406c9e6ae031c3dabecd1d8b398#diff-9cfca4a1b73e1e28e30fb9b0b984aad6d4caaf0819c61ed40ad338600531f745R327 which was part of the addition of multi-protocol as an optional feature, in the PR that fixed #56.

So I think:

* a Receiving Server that is able to process and apply a `URI` for WebDAV SHOULD announce this
* a Sending Server SHOULD NOT send a `URI` unless the Receiving Server has announced it understands it

### LINKED ISSUE #133: Rethink capability name `/mfa-capable`
We can change the string `/mfa-capable` to something without a leading slash. 
Also, the 'capable' is already implied by the fact that this string appears in a list of capabilities
And maybe we want to add _what_ the server is capable of doing. It's not only capable of MFA, we want it to be able to securely and reliably enforce an MFA requirement on a share.
So maybe something like 'mfa-requirement' or 'enforce-mfa-requirements'?

michielbdejong:
In my presentation about OCM at CS3 2023 in Barcelona I had this slide about CAPabilitieS DISCOvery ;)
cs3-2023-bcn-ocm 001

Now I'm thinking if we put good capability discovery into the `/.well-known/ocm` / `/ocm-provider` document, then discovery doesn't even have to rely on the `apiVersion` field we have there. Older implementations will just have less entries in the capabilities object, or the object will be missing from the document altogether.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -161,8 +161,8 @@ The Invite Acceptance Response SHOULD be a HTTP response:
   * REQUIRED: `email` - non-normative / informational; an email address for the Invite Sender. Not necessarily at the same FQDN as their OCM Server
   * REQUIRED: `name` - human-readable name of the Invite Sender, as a suggestion for display in the Invite Receiver's address book
 
-A 200 response status means the Invitation Acceptance Request was successful.
-A 400 response status means the Invitation Token is invalid or does not exist.
+A 200 response status means the Invite Acceptance Request was successful.
+A 400 response status means the Invite Token is invalid or does not exist.
 A 403 response status means the Invite Receiver OCM Server is not trusted to accept this Invite.
 A 409 response status means the Invite was already accepted.
 
@@ -234,7 +234,7 @@ Step 7: The JSON response body is the data that was discovered.
 The JSON response body offered by the Discoverable Server SHOULD contain the following information about its OCM API:
 
 * REQUIRED: enabled (boolean) - Whether the OCM service is enabled at this endpoint
-* REQUIRED: apiVersion (string) - The OCM API version this endpoint supports. MUST start with `"1."` and clients MUST ignore the rest of the string.
+* REQUIRED: apiVersion (string) - The OCM API version this endpoint supports. MUST start with `"1."` for this version of the specification and clients MUST ignore the rest of the string.
 * REQUIRED: endPoint (string) - The URI of the OCM API available at this endpoint. Example: `"https://my-cloud-storage.org/ocm"`
 * OPTIONAL: provider (string) - A friendly branding name of this endpoint. Example: `"MyCloudStorage"`
 * REQUIRED: resourceTypes (array) - A list of all resource types this server supports in both the Sending Server role and the Receiving Server role, with their access protocols. Each item in this list should
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -278,6 +278,7 @@ itself be an object containing the following fields:
           Example: `["receive-code", "webdav-uri"]`. The array MAY include for instance:
     * `"enforce-mfa"` - to indicate that this OCM server can apply a Sending Server's MFA requirements for a Share on their behalf.
     * `"webdav-uri"` - to indicate that this OCM server can append a relative URI to the path listed for WebDAV in the appropriate `resourceTypes` entry
+    * `"protocol-object"` - to indicate that this OCM server can receive a Share Creation Notification whose `protocol` object contains one property per supported protocol instead of containing the standard `name` and `options` properties.
     * `"receive-code"` - to indicate that this OCM server can receive a `code` as part of a Share Creation Notification, and exchange it for a bearer token at the Sending Server's `/token` API endpoint.
 * OPTIONAL: criteria (array of string) - The criteria for accepting a Share Creation Notification.
           As all Receiving Servers should require the use of TLS in API calls,
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -234,10 +234,10 @@ Step 7: The JSON response body is the data that was discovered.
 The JSON response body offered by the Discoverable Server SHOULD contain the following information about its OCM API:
 
 * REQUIRED: enabled (boolean) - Whether the OCM service is enabled at this endpoint
-* REQUIRED: apiVersion (string) - The OCM API version this endpoint supports. Example: `"1.1.0"`
+* REQUIRED: apiVersion (string) - The OCM API version this endpoint supports. MUST start with `"1."` and clients MUST ignore the rest of the string.
 * REQUIRED: endPoint (string) - The URI of the OCM API available at this endpoint. Example: `"https://my-cloud-storage.org/ocm"`
 * OPTIONAL: provider (string) - A friendly branding name of this endpoint. Example: `"MyCloudStorage"`
-* REQUIRED: resourceTypes (array) - A list of all supported resource types with their access protocols. Each item in this list should
+* REQUIRED: resourceTypes (array) - A list of all resource types this server supports in both the Sending Server role and the Receiving Server role, with their access protocols. Each item in this list should
 itself be an object containing the following fields:
   * name (string) -  A supported resource type (file, folder, calendar, contact, ...).
                 Implementations MUST support `file` at a minimum. Each resource type is identified by its `name`: the list MUST NOT
@@ -246,7 +246,7 @@ itself be an object containing the following fields:
                 The supported recipient share types.
                 MUST contain `"user"` at a minimum, plus optionally `"group"` and `"federation"`.
                 Example: `["user"]`
-  * protocols (object) - The supported protocols for accessing shared resources.
+  * protocols (object) - The supported protocols for accessing shared resources of this type.
                 Implementations MUST support at least `webdav` for `file` resources,
                 any other combination of resources and protocols is optional. Example:
                 ```json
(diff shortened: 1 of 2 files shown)


michielbdejong:
The current `resourceTypes` object is weird because it seems to be used both to advertise what types of resources a server can receive, and to advertise details about how this server acts as a resource server.

glpatcern:
Concerning apiVersion, IMHO its useful for humans, as the branding, and should NOT be relied upon to decide on caps. Yet it is used for that by Nextcloud.

michielbdejong:
Capabilities should be about being able to process something you receive.
Being able to send something or not is not so interesting to discover.

README.md
```diff
@@ -236,10 +236,10 @@ Step 7: The JSON response body is the data that was discovered.
 The JSON response body offered by the Discoverable Server SHOULD contain the following information about its OCM API:
 
 * REQUIRED: enabled (boolean) - Whether the OCM service is enabled at this endpoint
-* REQUIRED: apiVersion (string) - The OCM API version this endpoint supports. Example: `"1.1.0"`
+* REQUIRED: apiVersion (string) - The OCM API version this endpoint supports. MUST start with `"1."` and clients MUST ignore the rest of the string.
```
glpatcern: I'd not impose restrictions here on the content. Maybe just `It is provided for information purposes only: clients SHOULD NOT infer capabilities based on its value`.

glpatcern (commented):
Looks good, I think we can include the full list in the spec as per #121 !

glpatcern:
@michielbdejong can you please rebase this on top of `develop`, as opposed to `caps-disco-base` ?

michielbdejong:
Rebased on develop.

michielbdejong:
I think we want to start off our capabilities list with all functionalities that we introduces since 1.0-proposal1 and for which one server needs to know whether the other one supports it. I can think of the following list:
* ability to receive a federation-type share (but this should go into the share types object)
* ability to receive a share without a 'protocol.name' being set to 'webdav' and without 'protocol.options' being present
* ability to receive a share with more than one protocol (maybe merge this with the previous one?)
* ability to receive a share with a webdav uri other than the base uri from discovery (do we think this is useful, actually?)
* ability to use the code flow to obtain a bearer token (i.e. receive shares that do not have a sharedSecret)
* ability to honour an MFA requirement (this is a super important one, we don't want requirements to get lost in translation!)

Functionalities which I think don't make sense to announce (these functionalities can just be present unannounced):
* /ocm-provider being copied to /.well-known/ocm (it makes no sense to announce that in the doc itself)
* ability to send a federation share (this is implicit when such a share is sent)
* ability to send a share with an MFA requirement
* ability to send an invite (this is implicit in the existence of the invite)
* ability to receive an Invite Acceptance Request (this is implicit in the existence of the invite)
* ability to sign a http request

michielbdejong:
so caps to add:
* honour an MFA requirement
* receive a share with new 'protocol' format
* receive a share with a webdav uri that is a *relative* path to added to the discovered one (not an absolute URL)
* receive a share with a code (and no sharedSecret)

webapp URI should probably also somehow be bound to the server domain name from discovery

glpatcern:
Let's not forget:
* receive an invitation

As very first capability (in the form of endpoint)!

IETF-RFC.md
```diff
@@ -234,10 +234,10 @@ Step 7: The JSON response body is the data that was discovered.
 The JSON response body offered by the Discoverable Server SHOULD contain the following information about its OCM API:
 
 * REQUIRED: enabled (boolean) - Whether the OCM service is enabled at this endpoint
-* REQUIRED: apiVersion (string) - The OCM API version this endpoint supports. Example: `"1.1.0"`
+* REQUIRED: apiVersion (string) - The OCM API version this endpoint supports. MUST start with `"1."` and clients MUST ignore the rest of the string.
```
glpatcern: As already commented, I'd not enforce any value, in particular the `MUST start with 1.` - we will have 2.0 :)

Maybe just `It is provided for information purposes only: clients SHOULD NOT infer capabilities based on its value`, or (better) say that when discussing capabilities.

IETF-RFC.md
```diff
@@ -275,11 +275,20 @@ itself be an object containing the following fields:
...
-          Example: `["/notifications"]`. The array MAY include for instance:
-    * `"/notifications"` - to indicate this OCM server is capable of processing OCM Notifications
-    * `"/invite-accepted"` - to indicate that this OCM server is capable of processing Invite Acceptance Requests.
-    * `"/mfa-capable"` - to indicate that this OCM server can apply a Sending Server's MFA requirements for a Share on their behalf.
-        
+          Example: `["receive-code", "webdav-uri"]`. The array MAY include for instance:
+    * `"enforce-mfa"` - to indicate that this OCM server can apply a Sending Server's MFA requirements for a Share on their behalf.
+    * `"webdav-uri"` - to indicate that this OCM server can append a relative URI to the path listed for WebDAV in the appropriate `resourceTypes` entry
+    * `"receive-code"` - to indicate that this OCM server can receive a `code` as part of a Share Creation Notification, and exchange it for a bearer token at the Sending Server's `/token` API endpoint.
+* OPTIONAL: criteria (array of string) - The criteria for accepting a Share Creation Notification.
```
glpatcern, line 284: Introducing `criteria` in itself is an interesting concept, though I have the feeling that it complements - but also quite overlaps? - with the `requirements` for shares.

With a "`A` shares to `B`" scenario:
- A `requirement` is a prescription that `A` asks `B` to respect when accessing `A`'s resource
- A `criterium` is a restriction `B` implements in accepting `A`'s sharing of a resource, IIUC.

So the consequence would be that `A`'s system is expected to check `B`'s discovery endpoint prior to proposing `A`'s users to share to one of `B`'s users. Right? Or otherwise, where/when is this information going to be consumed?

glpatcern (commented):
A few preliminary comments for some discussion

IETF-RFC.md
michielbdejong: My idea was that servers can announce which version of the spec they support. So if they support the version that says "announce me as 1.*", then that is what they should announce.

We could keep it as "anything goes" but we don't want a 1.* server to announce a 2.* API version, right?

> we will have 2.0 :)

Exactly, and then we'll change the "MUST start with 1.*" to "MUST start with 2.*", that's how we link the api version to the spec text. We can discuss it in today's meeting!

IETF-RFC.md
michielbdejong: Yes!

`A`'s system could check if it can satisfy all the criteria of `B`'s system, although of course with allowlist and denylist, it can never be sure. If it knows that the share will fail, it can make the UX nicer by not offering it as an option in the first place.

Another way to use this would be for debugging.

michielbdejong:
>  Let's not forget:
> receive an invitation
> As very first capability (in the form of endpoint)!

There are two sides to this and I think neither of them warrant a discoverable capability. When Alice sends an invite to Bob:

* capability to receive an invitation - this is between Bob and Bob's server, not something Alice's server would, or even could discover. Remember, if Alice's server knew the FQDN of Bob's server then the invite would be unnecessary.
* capability to receive a server-to-server [invite acceptance](https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md#invite-acceptance-request-details) request - if Alice's server allows Alice to mint invites containing its FQDN, then that in itself already implies that it is capable of also receiving acceptance notifications for it, right?

IETF-RFC.md
glpatcern: (I'll miss today's meeting)
Hmm I'm not sure - IMHO the spec's version is an attribute, a "payload", not an inherent feature ("MUST start with..."). 

And whether a server announces version `A.B` but actually implements `C.D`, this is more a matter of compliance that can (and should) be independently tested, e.g. by (y)our test suite, not of "adhering to a standard" I think.

IETF-RFC.md
glpatcern: Understood! Then it's probably good to include that in the spec?

mickenordin (approved):
LGTM

glpatcern:
> * Remember, if Alice's server knew the FQDN of Bob's server then the invite would be unnecessary.

Are you sure? The principle of invite is for a user to add a trusted party, even if the Server URL of the other party is well known. I think discovering whether B's server supports acceptance of invites (which ok implies also issuance of invites, both being an optional feature, and thus a capability), even without a WAYF page, is necessary for A's server to expose the invite flow altogether.

IETF-RFC.md
```diff
@@ -489,7 +499,7 @@ Otherwise, if `protocol.webdav.sharedSecret` is not empty, the receiver MUST pas
 
 In both cases, when the Resource is a folder and the Receiving Server accesses a resource within that shared folder, it SHOULD append its relative path to that URL.
 
-Additionally, if `protocol.<protocolname>.requirements` include `mfa-enforced`, the Receiving Server MUST ensure that the Receiving Party has been authenticated with MFA.
+Additionally, if `protocol.<protocolname>.requirements` includes `mfa-enforced`, the Receiving Server MUST ensure that the Receiving Party has been authenticated with MFA.
```
glpatcern, line 502:
```suggestion
Additionally, if `protocol.<protocolname>.requirements` includes `enforce-mfa`, the Receiving Server MUST ensure that the Receiving Party has been authenticated with MFA.
```

MahdiBaghbani (approved):

mickenordin:
Ok, my understanding is that the two points from @glpatcern is that:

1. we should NOT do the change on line 237, that is not change the wording around versioning.
2. We should leave invitations in as a capability.

For me I am fine with both these modifications to this pr, and if we do them, we can go ahead and merge this pr which would be a great step onwards to version 1.2. So what do you say @michielbdejong can you agree to those two changes?

glpatcern (approved):
Looks good! Do you want to patch the `spec.yaml` in this PR as well?

michielbdejong:
Checked that it looks OK on [spec.yaml render of this branch](https://cs3org.github.io/OCM-API/docs.html?branch=caps-disco&repo=OCM-API&user=cs3org):
Screenshot 2024-11-14 at 13 49 35

PR #144, opened 2024-09-29
Small fix for code spec

MahdiBaghbani:
Additionally, we should also mark all fields as required in the spec.

I think we haven't included this endpoint on the Internet-Draft yet and left it for OCM 2.0, right?


glpatcern (approved):
This looks correct, and my understanding is that the token exchange flow will be part of 1.2, without waiting for 2.0!

michielbdejong (approved):

PR #143, opened 2024-09-29
Explicitly mark invite flow attributes as required

MahdiBaghbani:
This PR makes the `name` and `email` fields in the invite flow as optional and everything else is required.

Fixes #142
Fixes #117

PR #148, opened 2024-10-10
Explicitly mark invite flow attributes as required

MahdiBaghbani:
This is #143 with good diff.


glpatcern (approved):
Yep, this is good now, even though the commits are a bit weird (but we can squash+merge).

PR #156, opened 2024-10-17
use-code as new requirement + cleaned up superseded references to mfa-capable

glpatcern:
This is a draft to extend `requirements` along the concept that "a requirement is an enforced capability that otherwise would be optional".

Proper definition of capabilities is required prior to merge this.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -402,9 +402,13 @@ If `multi` is given, one or more protocol
                       - `read` allows read-only access including download of a copy.
                       - `write` allows create, update, and delete rights on the resource.
                       - `share` allows re-share rights on the resource.
-                      - `mfa-enforced` requires the user accessing the resource to be
-                        MFA-authenticated. This permission MAY be used if the
-                        provider exposes the `mfa-capable` capability.
+     * OPTIONAL requirements (array of strings) -
+                      The requirements that the sharee MUST fulfill to access the resource. A subset of:
+                      - `mfa-enforced` requires the consumer to be MFA-authenticated. This
+                        MAY be used if the recipient provider exposes the `enforce-mfa` capability.
+                      - `use-code` requires the consumer to exchange the given `code` via a
+                        signed HTTPS request. This MAY be used if the recipient provider exposes
+                        the `receive-code` capability.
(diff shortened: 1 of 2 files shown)


glpatcern:
Following today's discussion, an idea to introduce a `requirement` around HTTP signature would rather be with the following flow:
1. User A shares a resource to recipient B by:
* Making a signed request to B's endpoint (that's actually optional, but would be fair and nice for A to do)
* Including a `code` in the payload
* Setting an empty `sharedSecret`
2. User B is then expected to access A's resource by:
* Exchanging the `code` via signed request to `/token` at A
* Executing `PROPFIND` etc. against A's resource with the short-lived token, without any further signature

Does that make sense?

If so, IMHO we should explicitly define a `requirement` to enforce that user B go with this flow. Implicitly assuming it in the case `sharedSecret` is empty does not look satisfactory...

glpatcern:
Following today's discussion I have rephrased the specification of the requirement for a signed request. Still, this PR is to be considered after #136 about Caps Discovery.

michielbdejong:
@glpatcern IIUC this is now covered by the criteria in [discovery](https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md#fields), correct? So then we can close this?

glpatcern:
The `criteria` would apply to _all_ shares from a given provider, therefore they're admin-driven. Whereas here we would introduce a requirement that applies to a specific share, and it might be given as an option to the user.

michielbdejong:
Ah wait, I understand now. I thought this was about the signature on the Share Creation Notification. But it refers to the signature used when exchanging the code for a short-lived access token, right?

As you say, merely because a code is present in a share and a sharedSecret is not, is a bit of a weak way to indicate that the recipient is required to use the code flow.

So maybe we can call this requirement "use-code"?

glpatcern:
> So maybe we can call this requirement "use-code"?

Makes sense, yes. I'll do that, and I'll also drop the `/mfa-capable` "fake" endpoint.

mickenordin (commented):
Good job!

mickenordin (approved):

michielbdejong (approved):

PR #158, opened 2024-10-22
Improve share access process and share examples

### LINKED ISSUE #157: Path-to-resource examples in Share endpoint
Following today's meeting, I realize my own confusion with the paths given as examples in the `/share` endpoint, which should be corrected:

* When sending off a share, the path should be to the shared resource, nothing more, that is for example `/dav/ocm/<some-uuid>` (already dropping the host here).
* The receiver is expected to compose the URL as in OCM 1.0, and MAY issue requests such as `PROPFIND https://<webdav-fqdn-as-advertised-in-discovery>/dav/ocm/<some-uuid>/some/relative/path/to/resource.txt`, with a bearer token
* The sharing end is expected to support such "targeted" HTTP requests

I propose to amend the example to clarify that.

And at this point... We may well get closer to the OCM 1.0 principle, where the share payload contains just a `sharedKey` and a `sharedSecret` (not just the secret - the key here is really for lookup, so it ends up being logged, and the secret is the bearer token, with the option of a `code`), and the receiver is expected to compose the URL as:

`https://<discovered-fqdn>/<webdav-root-from-discovery>/<sharedKey>/relative/path/to/resource.txt`

This should simplify the ongoing work with Nextcloud, and it requires a minor amendment of Reva where we have to issue a discovery request (which we MUST do for security reasons) and store in the DB the full URL as opposed to just having it ready in the payload.

I confess I'd prefer `shareKey` instead of `sharedKey`, if we were to go this way, but we already have `sharedSecret`...

@michielbdejong @mickenordin @MahdiBaghbani comments?

glpatcern:
This PR is an attempt to fix #157, but in doing so it partly breaks the v1.1 protocol to access a share.

In particular, for the `webdav` and `datatx` protocols compatibility is ensured by keeping the same property name `uri` (resp. `srcUri`), but the recommended value is now a simple identifier or "key", not a full URI.

For the `webapp` protocol I propose to break compatibility, assuming that the only implementation that requires to be adapted is Reva.

This is still work in progress as I'm myself not fully happy, the main outstanding question being shall we actually drop `uri` and go for `sharedKey`?

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -186,7 +186,7 @@ Second, when the Receiving Party accepts the Invite, the Receiving Server knows 
 
 Third, equivalently, the Sending Server knows it is essentially registering the Receiving Server as an API client at the request of the Receiving Party, to whom the right to request this has been traceably delegated by the Sending Party, which is one of its registered users.
 
-Fourth, related to the second one, it removes the partial 'open relay' problem that exists when the Sending Server is allowed to include any Receiving Server FQDN in the Sending Gesture. Without the use of Invites, a Distributed Denial of Service attack could be organised if many internet users collude to flood a given OCM Server with Share Creation Notifications which will be hard to distinguish from legitimate requests without human interaction. An unsolicited (invalid) Invite Acceptance Request is much easier to filter out than an unsolicited (possibly valid, possibly invalid) OCM request, since the Invite Acceptance Request needs to contain an Invite Token that was previously uniquely generated at the Invite Sender OCM server.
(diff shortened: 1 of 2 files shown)


mickenordin (approved):
I think breaking backwards compatibility for webapps should be ok, the one I was thinking about except reva was Nextcloud talk, but they don't use webbapp, but rather talk-v1:

```
    {
      "name": "talk-room",
      "shareTypes": [
        "user"
      ],
      "protocols": {
        "talk-v1": "/ocs/v2.php/apps/spreed/api/"
      }
    }
 ```

glpatcern:
That's right, `talk` is a "protocol" on its own, and I can obviously take responsibility to align Reva's `webapp` support to OCM 1.2 if we go ahead with this PR.

michielbdejong (approved):
LGTM

michielbdejong:
I think it's OK to leave it as "uri", with the note that it SHOULD be a relative one.

PR #160, opened 2024-11-14
Mark reshare notification as out of scope

### LINKED ISSUE #159: TODOs in the protocol specification
I came across the following TODOs in the IETF specification, that we should address prior to publication:

```
# Share Updating
TODO: document `"RESHARE_CHANGE_PERMISSION"`

# Resharing
[...]
TODO: document how the Receiving Party can know if the Sending Party understood and processed the
reshare request.
```

michielbdejong:


glpatcern (approved):
Looks good, I agree to mark them as out of scope at this point, also given that only one implementation has "something".

mickenordin (approved):
LGTM

PR #165, opened 2025-02-25
Clarify how access happens for shares without absolute URIs

glpatcern:
As we are implementing OCM 1.2 in Reva, we realized that full URIs in new share payloads are:
1) Not less trustworthy than any other metadata a receiver can use to establish the sender's address.
2) ~~Actually required~~ if we want to support deployments that make use of reverse proxies (such as CERNBox), where the receiver has no other way than contacting the sender at their advertised URI (typically different from the sender's address). Edit: if the owner's OCM address is provided, then full URIs are not required.

See https://github.com/cs3org/reva/pull/5076#discussion_r1970426187 for more details.

With this PR, ~~we are "un-deprecating" the full URI concept in new share payloads~~ we make the OCM address' FQDN mandatory.

Additionally, some fixes and a re-harmonization of the IETF text have been included, most notably:
* The `webdav.uri` property is `REQUIRED`, not `OPTIONAL`
* Use of the `/.well-known/ocm` discovery endpoint
* No secret hashes in URIs
* `uriTemplate` was renamed to `uri`


mickenordin:
I think I am missing something.  Why is it not possible to use the endpoint designated in the discovery endpoint when using a reverse proxy?

Is it not the job of the reverse proxy to proxy the request to the correct backend server? It is not intended that the communication bypass the reverse proxy and go directly to the backend, right?

michielbdejong (changes requested):
I don't think I understand what the problem is. The way I understand the flow is:

* einstein@server1.com shares "files/bla.txt" with marie@server2.com
* server1.com does a lookup of server2.com/.well-known/ocm
* the Share Creation Notification is POSTed. This may go through proxies, doesn't matter. The important part here is part after the `@` symbol in the `sender` field, so in [our API docs example](https://cs3org.github.io/OCM-API/docs.html?branch=v1.2.0&repo=OCM-API&user=cs3org#/paths/~1shares/post) that would be `apiwise.nl` instead of server1.com.
* server2.com retrieves server1.com/.well-known/ocm
* checks whether it thinks this spam or not. It can do so for instance by checking the request headers against the pubkey announced there, or by checking whether the share is actually accessible at the announced WebDAV API of server1 (see next step).
* it uses the base path from [`resourceTypes[0].protocols.webdav`](https://cs3org.github.io/OCM-API/docs.html?branch=v1.2.0&repo=OCM-API&user=cs3org#/paths/https:~1~1%3Cdiscovery-fqdn%3E~1.well-known~1ocm/get) to find the webdav base URL, in our API docs example that would be `/remote/dav/ocm/` but this can also be an absolute URL at https://storage.server1.com or https://proxy.sever1.com or similar.
* It adds the relative path of the share to the WebDAV base URL to get (in this example) https://apiwise.nl/remote/dav/ocm/files/bla.txt as the WebDAV URL.

Which part of this is made impossible by the use of (reverse) proxies?

glpatcern:
OK, now I understand what you (both) meant: the key is all in `The important part here is part after the @ symbol in the sender field`. So effectively the spec says to use an information that is part of the payload, not in the request header / remote address (which cannot be trusted, that was the whole point).

I must say that I have missed that, I'll try and spell it out explicitly in the [discovery section](https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md#resource-access) even if it's explained in the Invite flow section.

At this point, no need to "un-deprecate" the full URL in the payload, on the basis that the payload MUST contain the server in the `owner` field. I will amend this PR to just leave the other corrections.

IETF-RFC.md
```diff
@@ -73,7 +73,7 @@ We define the following concepts (with some non-normative references to related
 * __Discovering Server__ - a server that tries to obtain information in OCM API discovery
 * __Discoverable Server__ - a server that tries to supply information in OCM API discovery
 * __OCM Address__ - a string of the form `<Receiving Party's identifier>@<fqdn>` which can be used to uniquely identify a user or group "at" an OCM Server. `<Receiving Party's identifier>` is an opaque string,
-unique at the server. `<fqdn>` is the Fully Qualified Domain Name by which the server is identified. This can, but doesn't need to be, the domain at which the OCM API of that server is hosted.
+unique at the server. `<fqdn>` is the Fully Qualified Domain Name by which the server is identified. This MUST be the domain at which the OCM API of that server is hosted.
```
glpatcern: Reviewing myself, I believe this is the only significant change, all the rest being a matter of better phrasing and realigning of the IETF text to the spec.

I believe we have to force a `MUST` here, otherwise my argument of un-deprecating full URIs in share payloads remains valid. Otherwise, if a share payload does not contain a usable owner, and the URI is relative, how can a receiver contact the sender, assuming that its remote address may not be valid?

michielbdejong: I think this "doesn't have to be" phrase refers to the fact we are allowing the OCM API  (i.e. where `POST ./shares` and other endpoints live) as advertised in `/.well-known/ocm` to be on a different server than `/.well-known/ocm` itself, right?

glpatcern: Ah, my interpretation was that the `/.well-known/ocm` must be on the same server as the owner's (or recipient's) OCM address that is sent by a `POST ./shares`

michielbdejong:
```suggestion
unique at the server. `<fqdn>` is the Fully Qualified Domain Name by which the server is identified. This MUST be the domain at which the `/.well-known/ocm` endpoint of that server is hosted.
```

IETF-RFC.md
```diff
@@ -492,7 +489,7 @@ They could give the Receiving Party the option to accept or reject the Share, or
 
 # Resource Access
 To access the Resource, the Receiving Server MAY use multiple ways, depending on the body of the Share Creation Notification. The procedure is as follows:
-1. A root path `<sender-ocm-path>` MUST be obtained by querying the [Discovery](#ocm-api-discovery) endpoint at the Sending Server and getting `resourceTypes[0].protocols.webdav`.
+1. The receiver MUST extract the OCM Server FQDN from the `owner` field of the received share, and MUST query the [Discovery](#ocm-api-discovery) endpoint at that address: the `resourceTypes[0].protocols.webdav` value is the `<sender-ocm-path>` to be used in step 3.
```
michielbdejong: `sender`, not `owner`, right?
```suggestion
1. The receiver MUST extract the OCM Server FQDN from the `sender` field of the received share, and MUST query the [Discovery](#ocm-api-discovery) endpoint at that address: the `resourceTypes[0].protocols.webdav` value is the `<sender-ocm-path>` to be used in step 3.
```

IETF-RFC.md
```diff
@@ -394,6 +394,10 @@ If `multi` is given, one or more protocol
                 only support `webdav`.
 
    * Protocol details for `webdav` MAY contain:
+     * REQUIRED uri (string)
```
michielbdejong, line 397: why is this one moving from optional to required, was that discussed?

IETF-RFC.md
```diff
@@ -410,19 +414,12 @@ If `multi` is given, one or more protocol
...
-                    MAY include a secret hash in the path, in which case there MAY be
-                    no associated `sharedSecret`.
    * Protocol details for `webapp` MAY contain:
-     * REQUIRED uriTemplate (string)
-                    A templated URI to a client-browsable view of the shared resource,
-                    such that users may use the web applications available at the site.
-                    The URI MAY include a secret hash in the path. If the path includes
-                    a `{relative-path-to-shared-resource}` placeholder, implementations
-                    MAY replace it with the actual path to ease user interaction.
+     * REQUIRED uri (string)
```
michielbdejong, line 418: is this a breaking change?

michielbdejong (commented):
see inline comments

IETF-RFC.md
glpatcern: In fact it has always been required (in spec.yaml), the IETF text was actually incorrect

IETF-RFC.md
glpatcern: (see previous comment, it's not a change)

IETF-RFC.md
```diff
@@ -224,17 +224,17 @@ At the start of the process, the Discovering Server has either an OCM Address, o
 
 Step 1: In case it has an OCM Address, it should first extract `<fqdn>` from it (the part after the `@` sign).
```
```suggestion
Step 1: In case it has an OCM Address, it should first extract `<fqdn>` from it (the part after the last `@` sign).
```

mickenordin (approved):
LGTM

michielbdejong (approved):

PR #181, opened 2025-05-31
GitHub action for Spec validation

glpatcern:
This PR adds an action to validate the spec against the OpenAPI standard.

The validation currently [fails](https://github.com/glpatcern/OCM-API/actions/runs/15361456386/job/43229085223), and we should try and address what does not respect the OpenAPI standard. In particular, it appears that `patternProperties` are not allowed, despite they are advertised as valid.

Edit: the PR was brought to successfully pass the validation, but meanwhile #182 brings additional fixes and will be based on top of this PR.

### DIFF
--- a/.github/workflows/oas-validation.yml
+++ b/.github/workflows/oas-validation.yml
@@ -1,3 +1,4 @@
+name: Validate OpenAPI Specification
 on: [push]
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/oas-validation.yml
```diff
@@ -0,0 +1,14 @@
+on: [push]
```
MahdiBaghbani, line 2: Cosmetic proposal for better view of the workflow in the Actions tab
```suggestion
name: Validate OpenAPI Specification
on: [push]
```

MahdiBaghbani:
I think we both did the same thing 😄

PR #182, opened 2025-05-31
fix: OpenAPI Specification

MahdiBaghbani:
Related to PR #181 
- Added `servers` section to remove the URL parts before the `/` in the `paths`, Also beneficial when testing a live EFSS by simply entering its base URL and then playing with the commands in Swagger UI
Screenshot 2025-05-31 171238
Screenshot 2025-05-31 171252

- Fixed the `patternProperties` by removing it and moving the known parts under protocols and optional arbitrary protocols under `additionalProperties`, seems to fix the swagger validator
Screenshot 2025-05-31 170900

- Added an example for RO-Crate as an optional protocol
Screenshot 2025-05-31 170809

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -1,11 +1,4 @@
 openapi: 3.0.3
-x-origin:
-  - url: https://raw.githubusercontent.com/cs3org/OCM-API/develop/spec.yaml
-    format: swagger
-    version: "2.0"
-    converter:
-      url: https://github.com/mermade/oas-kit
-      version: 7.0.8
 info:
   title: Open Cloud Mesh API
   description: Open Cloud Mesh Open API Specification.
--- a/spec.yaml
+++ b/spec.yaml
@@ -5,6 +5,11 @@ info:
   version: 1.2.0
   x-logo:
     url: logo.png
+servers:
+  - url: https://{discovery_fqdn}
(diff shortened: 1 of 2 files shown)


mickenordin (approved):
LGTM

glpatcern:
Merging #181 first and then this one

glpatcern:
Hm GitHub didn't understand that this PR was based on top of #181, @MahdiBaghbani could you please rebase?

Also, I have a comment on the `talk` example, if you didn't figure out the payload used by Nextcloud maybe better to remove it. The current example does not seem right (a talk does not have `read`+`write` permission)

spec.yaml
```diff
@@ -575,158 +572,156 @@ components:
...
+                    The size of the file to be transferred from the sending
+                    server.
+
+          additionalProperties:
+            type: object
+            description: >
+              Any optional additional protocols supported for this resource
+              MAY
+              be provided here, along with their custom payload. Appropriate
+              capabilities MUST be advertised in order for a sender to ensure
```
glpatcern: Bringing back this from #181:
```suggestion
              capabilities SHOULD be advertised in order for a sender to ensure
```

glpatcern (approved):

PR #184, opened 2025-06-02
WAYF Page: Add description of the network JSON Document

mickenordin:
This patch descripbes the format and utility of the network JSON Document, that is a fixed list of servers that can be used as a basis for discovery on the wayf page.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -129,7 +129,7 @@ OCM Servers MAY enforce a policy to only accept Shares between such trusted cont
 
 * the Invite Sender OCM Server generates a unique Invite Token and helps the Invite Sender to create the Invite Message
 * the Invite Sender uses some out-of-band communication to send the Invite Message, containing the Invite Token and the Invite Sender OCM Server FQDN, to the Invite Receiver
-* the Invite Receiver navigates to the Invite Receiver OCM Server (possibly using a Where-Are-You-From page provided as part of the Invite Message) and makes the Invite Acceptance Gesture
+* the Invite Receiver navigates to the Invite Receiver OCM Server (possibly using a WAYF Page provided as part of the Invite Message) and makes the Invite Acceptance Gesture
 * the Invite Receiver OCM Server discovers the OCM API of the Invite Sender OCM Server using generic OCM API Discovery (see section below)
 * the Invite Receiver OCM Server sends the Invite Acceptance Request to the Invite Sender OCM Server
 
@@ -325,7 +325,7 @@ itself be an object containing the following fields:
             Example: https://my-cloud-storage.org/ocm#signature
   * REQUIRED publicKeyPem (string) - PEM-encoded version of the public key.
             Example: "-----BEGIN PUBLIC KEY-----\nMII...QDD\n-----END PUBLIC KEY-----\n"
-* OPTIONAL: inviteAcceptDialog (string) - URL path of a web page where a user can accept an invite, when query parameters `"token"` and `"providerDomain"` are provided. Implementations that offer the `"invites"` capability SHOULD provide this URL as well in order to enhance the UX of the Invite Flow. If for example `"/index.php/apps/sciencemesh/accept"` is specified here then a Where-Are-You-From page could redirect the end-user to `/index.php/apps/sciencemesh/accept?token=zi5kooKu3ivohr9a&providerDomain=example.com`.
+* OPTIONAL: inviteAcceptDialog (string) - URL path of a web page where a user can accept an invite, when query parameters `"token"` and `"providerDomain"` are provided. Implementations that offer the `"invites"` capability SHOULD provide this URL as well in order to enhance the UX of the Invite Flow. If for example `"/index.php/apps/sciencemesh/accept"` is specified here then a WAYF Page SHOULD redirect the end-user to `/index.php/apps/sciencemesh/accept?token=zi5kooKu3ivohr9a&providerDomain=example.com`.
 
 # Share Creation Notification
 To create a share, the sending server SHOULD make a HTTP POST request
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -157,7 +157,7 @@ If the inviteAcceptDialog uses a WAYF page, it MAY consist of a fixed list of se
   * REQUIRED: `url` - the OCM Server's FQDN.
   * REQUIRED: `displayName` - a human-readable name for the OCM Server.
 
-Implementors that provide a WAYF Page SHOULD make the URL for the API endpoint of such a central database configurable, allowing the OCM Server to be part of a network of associated OCM Servers. The configuration mechanism MAY allow an OCM Server to be part of multiple networks, thus displaying a union of multiple lists in it's WAYF Page.
+Implementors that provide a WAYF Page SHOULD make the URL for the API endpoint of such a central database configurable, allowing the OCM Server to be part of a network of associated OCM Servers. The configuration mechanism MAY allow an OCM Server to be part of multiple networks, thus displaying a union of multiple lists in its WAYF Page.
 
 ### Invite Acceptance Response Details
 The Invite Acceptance Response SHOULD be a HTTP response:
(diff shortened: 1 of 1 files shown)


IETF-RFC.md
```diff
@@ -151,6 +152,13 @@ Since the Invite Flow does not require either Party to type or remember the `use
 
 Also, a different `userID` could be given out to each contact, to avoid correlation of identities.
 
+If the inviteAcceptDialog uses a WAYF page, it MAY consist of a fixed list of servers, in addition to, or instead of, a free-text input where any OCM server can be entered. This is especially useful if the Invite Sender is part of a network of associated OCM Servers. The Invite Sender's server MAY use a central database to store the fixed list of OCM servers. In that case the central database MUST be exposed as a JSON document.
```
glpatcern: I'd replace `database` -> `directory`. Also, shall we add that: the Directory MUST offer an HTTP endpoint, such that a `HTTP GET` would return a JSON document with the following fields...

IETF-RFC.md
```diff
@@ -151,6 +152,13 @@ Since the Invite Flow does not require either Party to type or remember the `use
 
 Also, a different `userID` could be given out to each contact, to avoid correlation of identities.
 
+If the inviteAcceptDialog uses a WAYF page, it MAY consist of a fixed list of servers, in addition to, or instead of, a free-text input where any OCM server can be entered. This is especially useful if the Invite Sender is part of a network of associated OCM Servers. The Invite Sender's server MAY use a central database to store the fixed list of OCM servers. In that case the central database MUST be exposed as a JSON document. 
+* The format of the JSON document must be an array of objects, with the following fields:
+  * REQUIRED: `url` - the OCM Server's FQDN.
+  * REQUIRED: `displayName` - a human-readable name for the OCM Server.
+
+Implementors that provide a WAYF Page SHOULD make the URL for the API endpoint of such a central database configurable, allowing the OCM Server to be part of a network of associated OCM Servers. The configuration mechanism MAY allow an OCM Server to be part of multiple networks, thus displaying a union of multiple lists in it's WAYF Page.
```
glpatcern: I like this, so we also capture the idea of being part of multiple federations.

Just a trivial fix:
```suggestion
Implementors that provide a WAYF Page SHOULD make the URL for the API endpoint of such a central database configurable, allowing the OCM Server to be part of a network of associated OCM Servers. The configuration mechanism MAY allow an OCM Server to be part of multiple networks, thus displaying a union of multiple lists in its WAYF Page.
```

IETF-RFC.md
redblom: The sentence 'If the inviteAcceptDialog uses a WAYF page,...' confuses me. I thought the starting point ís the WAYF page (service) which may try to discover ('use') the `inviteAcceptDialog` endpoint for each provider in the list. So in my thinking it's the other way around: the WAYF page 'uses' the `inviteAcceptDialog` endpoints. Following (clicking) one of the WAYF links will lead you, as receiver of the invite, to your `inviteAcceptDialog` on your EFSS instance.

MahdiBaghbani: My understanding of WAYF is also like @redblom

glpatcern: I agree this may be confusing, though in my previous comment I didn't interpret it this way.

Maybe it's worth describing the full flow in a scenario with an external WAYF service, as the [current Steps](https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md#steps) do only mention `possibly using a WAYF page...` without further details.

Also, I'm thinking to expand a bit the JSON format of the directory: @mickenordin may I add a commit to this feature branch? Or a PR?

glpatcern: OK, I gave it a try with a proposed expanded version, can you please have a look?
cc @mickenordin @redblom @MahdiBaghbani

IETF-RFC.md
```diff
@@ -151,6 +153,30 @@ Since the Invite Flow does not require either Party to type or remember the `use
 
 Also, a different `userID` could be given out to each contact, to avoid correlation of identities.
 
+If the Invite Sender OCM Server implements a WAYF Page, such a page MAY include a fixed list of servers, in addition to, or instead of, a free-text input where any OCM Server can be entered. This is especially useful if the Invite Sender is part of a federation of associated OCM Servers. In order to populate the list of associated OCM Servers, the Invite Sender's server MAY use a Directory Service: in that case, the Directory Service MUST expose, via anonymous HTTP GET, a JSON document with the following format:
+  * REQUIRED: `federation` - a human-readable name for the list of OCM Servers exposed by the Directory Service
+  * REQUIRED: `servers` - a JSON array of objects to describe the list of OCM Servers with the following string fields:
+    * REQUIRED: `url` - the OCM Server's FQDN
+    * REQUIRED: `displayName` - a human-readable name for the OCM Server
```
glpatcern: And now reviewing my own proposal: I'm not entirely sure this is the best way to specify the JSON format. Definitely we're not putting anything in `spec.yaml` for this as this is a third-party service: is that clear enough?

IETF-RFC.md
```diff
@@ -127,7 +129,7 @@ OCM Servers MAY enforce a policy to only accept Shares between such trusted cont
 
 * the Invite Sender OCM Server generates a unique Invite Token and helps the Invite Sender to create the Invite Message
 * the Invite Sender uses some out-of-band communication to send the Invite Message, containing the Invite Token and the Invite Sender OCM Server FQDN, to the Invite Receiver
-* the Invite Receiver navigates to the Invite Receiver OCM Server (possibly using a Where-Are-You-From page provided as part of the Invite Message) and makes the Invite Acceptance Gesture
+* the Invite Receiver navigates to the Invite Receiver OCM Server and makes the Invite Acceptance Gesture. This step MAY be facilitated if the Invite Sender OCM Server implements a WAYF Page, such that the Invite Message would include a link to it for the Invite Receiver to navigate to: the Invite Receiver would then be able to indicate their OCM Server and proceed with the Invite Acceptance Gesture without copying the Invite Token.
```
glpatcern: This is a proposal in response to https://github.com/cs3org/OCM-API/pull/184#discussion_r2135276151

IETF-RFC.md
```diff
@@ -276,11 +302,12 @@ itself be an object containing the following fields:
...
+    * `"webdav-uri"` - to indicate that this OCM Server can append a relative URI to the path listed for WebDAV in the appropriate `resourceTypes` entry
+    * `"protocol-object"` - to indicate that this OCM Server can receive a Share Creation Notification whose `protocol` object contains one property per supported protocol instead of containing the standard `name` and `options` properties.
     * `"invites"` - to indicate the server would support acting as an Invite Sender or Invite Receiver OCM Server. This might be useful for suggesting to a user that existing contacts might be upgraded to the more secure (and possibly required) invite flow.
-    * `"receive-code"` - to indicate that this OCM server can receive a `code` as part of a Share Creation Notification, and exchange it for a bearer token at the Sending Server's `/token` API endpoint.
+    * `"receive-code"` - to indicate that this OCM Server can receive a `code` as part of a Share Creation Notification, and exchange it for a bearer token at the Sending Server's `/token` API endpoint.
+    * `"invite-wayf"` - to indicate that this OCM Server exposes a WAYF Page to facilitate the Invite flow.
```
glpatcern, line 290: This is mostly a "nice to have", I don't think it's really useful for some discovery flow

MahdiBaghbani:
Rebase this branch on top of `develop` so it stop creating PR for every commit, currently this has previous version of the workflow which doesn't lock the PR only for the `develop` branch

mickenordin:
I think these additions were very good, much clearer!

IETF-RFC.md
MahdiBaghbani: I suggest relocating the entire Directory Service specification to a dedicated appendix and adding an example JSON format, since every WAYF page must be able to parse the Directory Service’s response without guessing, but we can't put it into `spec.yaml`

glpatcern: Makes sense!

MahdiBaghbani (approved):
LGTM ✅

glpatcern (approved):

PR #192, opened 2025-06-12
Relax minimum requirements to qualify as OCM Server

glpatcern:
This change is proposed, as discussed today, in the context of OCM implementations for third-party integrations (cf. #180 and #191).

The idea here is that the third-party component is not an EFSS, and may only offer limited support for a given type of resource, NOT including `file`. Yet, it would expose an OCM discovery endpoint and issue Share Creation Notification requests to recipients that possibly host a full-fledged OCM server.


mickenordin (approved):
Yes, this looks good according to our discussion today

PR #194, opened 2025-06-13
RequestShare: Introduce the notion of a share request

mickenordin:


spec.yaml
```diff
@@ -848,6 +896,22 @@ components:
...
+        - shareWith
+        - shareId
+      properties:
+        shareWith:
+          type: string
+          description: >
+            Consumer specific identifier of the user, group or federation that
+            wants to recive a share of the resource.
+          example: 51dc30ddc473d43a6011e9ebba6ca770@geant.org
+        shareId:
```
KrausMatthias: Is this the same as `providerId` in NewShare ?

IETF-RFC.md
```diff
@@ -454,6 +454,27 @@ The Receiving Server MAY discard the notification if any of the following hold t
...
 * an initial check shows that the resource cannot successfully be accessed through (any of) the protocol(s) listed
 
+# Request for a share
+If the Recieving Party knows of a resource that has not yet been shared, the Recieving Party MAY make a HTTP POST to request a share
+
+* to the `/request-share` path in the Sending Server's OCM API
+* using `application/json` as the `Content-Type` HTTP request header
+* its request body containing a JSON document representing an object with the fields as described below
+* using TLS
+* using [httpsig](https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12)
```
KrausMatthias: I guess that should be httpsig [RFC9421]

IETF-RFC.md
```diff
@@ -454,6 +454,27 @@ The Receiving Server MAY discard the notification if any of the following hold t
...
+
+## Fields
+
+* REQUIRED shareWith (string) 
+          Consumer specific identifier of the user, group or federation that
+          wants to recive a share of the resource.
```
```suggestion
          wants to receive a share of the resource.
```

IETF-RFC.md
```diff
@@ -454,6 +454,27 @@ The Receiving Server MAY discard the notification if any of the following hold t
...
+        Example: "51dc30ddc473d43a6011e9ebba6ca770@geant.org"
+* REQUIRED shareId (string)
```
glpatcern: Here we have a problem: a receiver won't know the (opaque) ID of a remote resource, but only a path (or something "human readable"?). Shall we name this simply `share`, or `sharePath`?

IETF-RFC.md
```diff
@@ -454,6 +454,27 @@ The Receiving Server MAY discard the notification if any of the following hold t
...
+If the Recieving Party knows of a resource that has not yet been shared, the Recieving Party MAY make a HTTP POST to request a share
+
+* to the `/request-share` path in the Sending Server's OCM API
+* using `application/json` as the `Content-Type` HTTP request header
+* its request body containing a JSON document representing an object with the fields as described below
+* using TLS
+* using [httpsig](https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12)
+
+## Fields
+
```
glpatcern, line 1032: Don't we also need the sender?
```suggestion
* REQUIRED sender (string)
          OCM Address of the user who will be requested to share the resource.
```

IETF-RFC.md
```diff
@@ -454,6 +454,27 @@ The Receiving Server MAY discard the notification if any of the following hold t
...
+* REQUIRED shareWith (string) 
+          Consumer specific identifier of the user, group or federation that
```
```suggestion
          OCM Address of the user, group or federation that
```

IETF-RFC.md
```diff
@@ -454,6 +454,27 @@ The Receiving Server MAY discard the notification if any of the following hold t
...
+          wants to recive a share of the resource.
+        Example: "51dc30ddc473d43a6011e9ebba6ca770@geant.org"
+* REQUIRED shareId (string)
+          A unique identifier for the resource.
+        Example: 1234567890abcdef
+
+After receiving a request for a Share, the Sending Party MAY send a Share Creation Notification to the Sending Party using the OCM address in the shareWith field.
```
glpatcern: Here I'd imagine the flow to be async: the server would respond HTTP 202 Accepted, signifying that the `sender` user was notified about the request to share. Once they accept, the normal Share Notification will take place.

glpatcern (changes requested):
I went for a review of this new flow, hope I have understood the rationale, and of course all open for further comments.

spec.yaml
```diff
@@ -36,6 +36,54 @@ paths:
...
+              $ref: "#/components/schemas/RequestShare"
+        description: The JSON object to request access to a share.
+        required: true
+      responses:
+        "201":
+          description: >
+            Consumer successfully received request.
+        "400":
+          description: >
+            Bad request due to invalid parameters, e.g. when `shareId` is missing.
```
glpatcern: If the `share` identifier/path is missing or not found, a 404 would be more appropriate here

glpatcern (commented):
Looks good! One thing I wonder before merging: `sender` might be confused with the sender of the `request-share`, which actually is the recipient or `shareWith`. Should we name it differently, such as `owner` ? Or the fact that we have the recipient makes it clear enough after all? `owner` has its own drawbacks for co-owned / co-administered resources...

mickenordin:
> Looks good! One thing I wonder before merging: `sender` might be confused with the sender of the `request-share`, which actually is the recipient or `shareWith`. Should we name it differently, such as `owner` ? Or the fact that we have the recipient makes it clear enough after all? `owner` has its own drawbacks for co-owned / co-administered resources...

Even though neither of them is perfect, I think that your proposal is the less imperfect one, so I changed it to owner :)

glpatcern (approved):

PR #196, opened 2025-06-13
ResourceDiscovery: Add a new resource discovery mechanism.

mickenordin:
The intention is to allow for a way to advertise the existence of resources at the OCM server.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -1992,8 +1992,8 @@ with the following format:
     * OPTIONAL: `rocrate` - an embedded JSON object
       following the [ROCRATE] data-entities
       specification.
-    * REQUIRED: `id` - the unique identifier of the
-      resource at the OCM Server
+    * REQUIRED: `providerId` - the unique identifier
+      of the resource at the OCM Server
   Example:
  ```json
  {
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -1973,12 +1973,15 @@ The complete changelog is updated in the OCM-API GitHub repository.
 * Added JSContact extension to IANA Considerations.
 * Changed example domain to use cloud.example.org per RFC 2606.
 
-# Appendix D: Resource Discovery Service
+# Appendix E: Resource Discovery Service
 
 An OCM Server MAY publicly advertise available resources.
 This is done via the `resourceAdvertismentUri`. It is
-expected to expose, via anonymous HTTP GET, a JSON document
-with the following format:
+expected to expose, via anonymous HTTPS GET, a signed JWS
+document [RFC7515], where the signing key MUST be made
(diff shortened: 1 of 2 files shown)


IETF-RFC.md
```diff
@@ -644,6 +645,83 @@ A third-party Directory Service is a back-end service used to federate multiple
...
+    * OPTIONAL: `displayName` - the human-readable name of the OCM Server
+    * OPTIONAL: `publicUrl` - an public URL that can be used for direct download via anonymous HTTP GET
+    * OPTIONAL: `rocrate` - an embedded JSON object following the [ROCRATE](https://www.researchobject.org/ro-crate/specification/1.1/data-entities.html) data-entities specification.
+    * REQUIRED: `id` - the unique identifier of the resource at the OCM Server
+  Example:
+ ```json
+ {
+   "server": "OCM Server 1",
+   "resources": [
+     {
```
KrausMatthias, line 2005: maybe individual resources could be aligned to protocol in Discovery? E.g. when a resource is only available via webdav, webapp or some other protocol?

How do I know if I need to send a ShareRequest as in #194 ?

MahdiBaghbani:
I tried fixing the merge conflict and line length, failed on the line length 😅

IETF-RFC.md
```diff
@@ -1203,6 +1205,94 @@ format:
...
+  * REQUIRED: `resources` - a JSON array of objects to describe
+    the list of OCM Servers with the following fields:
+    * OPTIONAL: `displayName` - the human-readable name of the
+      OCM Server
+    * OPTIONAL: `publicUrl` - an public URL that can be used for
+      direct download via anonymous HTTP GET
+    * OPTIONAL: `rocrate` - an embedded JSON object following the
+      [ROCRATE](https://www.researchobject.org/ro-crate/specification/1.1/data-entities.html)
+      data-entities specification.
+    * REQUIRED: `id` - the unique identifier of the resource at the OCM Server
```
MahdiBaghbani: It's nitpicking on my side, I think we should keep the names consistent across the endpoints.

The `providerId` is from [share creation](https://cs3org.github.io/OCM-API/docs.html?branch=v1.2.1&repo=OCM-API&user=cs3org#/paths/~1shares/post)

```suggestion
    * REQUIRED: `providerId` - the unique identifier of the resource at the OCM Server
```

KrausMatthias: isn't this wrong then according to #271 as the providerId will only be created when the share is sent?

I guess this would need to be some identifier from #196 ?

MahdiBaghbani: @KrausMatthias yes, I think it should have name like `resourceId` or something like that instead of `providerId`

and this is #196 :smile:

PR #197, opened 2025-06-13
update: docname of the IETF RFC in the markdown format

MahdiBaghbani:


glpatcern (approved):
Yes, I had this in my local branch :) didn't commit it because the submission didn't seem to have worked. Will look at it again next week.

PR #200, opened 2025-06-16
Changed definition of public key id to follow Nextcloud

glpatcern:
While checking the [discovery endpoint at Nextcloud](https://cloud.nextcloud.com/ocm-provider) for other purposes, I realized that they expose the public key id as `keyId`, not just `id`.

Since this is a relatively new entry in the discovery payload, and given that we are not aware of any other implementation that exposes it, I propose to change the standard to reflect the Nextcloud implementation. Yes, we are making a favor to Nextcloud!


glpatcern:
@ArtificialOwl this is for you ;-)

mickenordin (approved):
Sounds reasonable

PR #202, opened 2025-06-22
Rephrased headers of all endpoints to improve consistency

glpatcern:
This is mostly in preparation of the IETF presentation: we realized that the summary and description had a different style for each endpoint, mostly due to "history". This PR attempts to re-harmonize them and improve consistency.


mickenordin (approved):
Lgtm

PR #209, opened 2025-07-17
Use RFC link for httpsig

### LINKED ISSUE #205: Add a normative link to the httpsig RFC
I think when we say 'httpsig' in our internet draft, what we are referring to is [RFC 9421](https://datatracker.ietf.org/doc/rfc9421/) right?
We should probably add a normative link to it

glpatcern:


mickenordin (approved):
LGTM

MahdiBaghbani (approved):

KrausMatthias:
Hi @glpatcern 

rfc9421 uses 
* "Content-Digest" instead of "Digest", as header name of the digest, 
* "(request-target)" is now "@request-target"
* @authority might be used instead of host

maybe those should be adopted at https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md?plain=1#L1017 for consistency?

glpatcern:
> maybe those should be adopted at https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md?plain=1#L1017 for consistency?

Hi @KrausMatthias, thanks for raising that, we should indeed fix the example.

PR #215, opened 2025-07-20
Use correct BCP 14 boiler plate

mickenordin:
This is the correct format according to: https://www.rfc-editor.org/rfc/rfc8174.txt

Currently the author tools gives us a warning about this.


glpatcern (approved):
I see you're already active in taking benefit from the ongoing IETF 123!

PR #220, opened 2025-07-21
Try to add line breaks at 72 char and add stream

mickenordin:
Try to fix some of the linting issues


glpatcern (approved):
The only drawback is that the git history got almost wiped... :'-)

MahdiBaghbani:
@glpatcern yes, as if it's a completely a new file 🆕

MahdiBaghbani (approved):

PR #221, opened 2025-07-21
Add security consideration section

mickenordin:
This is required in an IETF standard, we should do a lot more work on it though.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -904,8 +904,8 @@ implement measures for auditing and adding only trusted servers into the
 discovery service.
 
 ### httpsig
-Using signed messages to verify that an OCM server is the server you
-think it is, is realy best practice, and should be done unless you have
+It is recommended to use signed messages to verify that an OCM server
+is the server you expect it to be, and should be done unless you have
 a niche use case. 
 
 ## Legacy shared secrets
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -886,6 +886,33 @@ knowing if the Sending Party understood and processed the reshare request or not
 # IANA Considerations
 This document has no IANA actions.
 
+# Security Considerations
+## Trust
+There are several areas that is not covered by this specification. Most
+importantly we do not provide a way of establishing trust between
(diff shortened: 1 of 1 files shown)


IETF-RFC.md
```diff
@@ -886,6 +886,33 @@ knowing if the Sending Party understood and processed the reshare request or not
 # IANA Considerations
 This document has no IANA actions.
 
+# Security Considerations
+## Trust
+There are several areas that is not covered by this specification. Most
```
```suggestion
There are several areas that are not covered by this specification. Most
```

IETF-RFC.md
```diff
@@ -886,6 +886,33 @@ knowing if the Sending Party understood and processed the reshare request or not
...
+importantly we do not provide a way of establishing trust between
+servers, even though some features of the protocol rely on trust, such
+as the mfa-enforced criteria.
```
```suggestion
as the `mfa-enforced` requirement.
```

IETF-RFC.md
```diff
@@ -886,6 +886,33 @@ knowing if the Sending Party understood and processed the reshare request or not
...
+know and have out of band connections with other users on an OCM server.
+
+Further more the Directory Service feature can be used to establish a
+trusted federation, where a central authority can be trusted to
+implement measures for auditing and adding only trusted servers into the
+discovery service.
+
+### httpsig
+Using signed messages to verify that an OCM server is the server you
+think it is, is realy best practice, and should be done unless you have
```
```suggestion
It is recommended to use signed messages to verify that an OCM server
is the server you expect it to be, and should be done unless you have
```

IETF-RFC.md
```diff
@@ -886,6 +886,33 @@ knowing if the Sending Party understood and processed the reshare request or not
...
+a niche use case. 
+
+## Legacy shared secrets
+Using the legacy format of shared secrets instead of short lived tokens
+is not a good idea.
```
```suggestion
The legacy format of an OCM Share Notification with shared secrets is
only provided for backwards compatibility with existing implementations.
Implementers SHOULD NOT use it and prefer short-lived tokens instead.
```

glpatcern (approved):
A few changes, otherwise it's good to go as a first draft

Issue #212, opened 2025-07-17
Resharing semantic and constraints


glpatcern:
The semantic of a re-share operation is to be further developed, in particular in the context of having introduced `requirements` and `capabilities`.

As discussed over Matrix, the case of an OCM share that is reshared as public link may pose issues when e.g. the share had an `enforce-mfa` requirement, as a public link would completely void that requirement.

In general, we should make it clear that the default behavior should follow the least privilege principle, and reshare is allowed only if explicitly granted by the originator of the share.

micbar:
I am just curious: Is resharing of received OCM shares possible?

glpatcern:
@micbar in fact it is - apparently in Nextcloud you can create a public link out of a received federated share. In Reva though we never implemented that, and I believe OpenCloud is the same.

IMHO we should not assume that resharing is possible / allowed.

PR #230, opened 2025-08-01
Add required sender field to notifications

KrausMatthias:


glpatcern:
The issue with adding a `required` field is that it's a breaking change with existing implementations. @mickenordin and @ArtificialOwl are best positioned to comment on this.

At a higher level, the `/notifications` endpoint is poorly specified - see e.g. #212 - and requires some care, without necessarily taking existing implementations as 100% correct.

KrausMatthias:
What do you think about explicitly recommending the use of httpsig to authenticate notifications instead? #234

mickenordin:
I agree that notifications are poorly specified, I can't think of a notification that does not involve a sender, so adding this will make it less poorly specified. I think we need this one, otherwise the notification does not make sense.

KrausMatthias:
If httpsig is used, the sender can be derived from that, if httpsig isn't used you can't trust the sender field here anyways.

PR #231, opened 2025-08-01
Fixed HTTPSig example + removed duplicate in ack section

### LINKED ISSUE #228: Fix example for HTTP Sig
rfc9421 uses 
* "Content-Digest" instead of "Digest", as header name of the digest, 
* "(request-target)" is now "@request-target"
* @authority might be used instead of host

maybe those should be adopted at https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md?plain=1#L1017 for consistency?

_Originally posted by @KrausMatthias in https://github.com/cs3org/OCM-API/issues/209#issuecomment-3119425270_

glpatcern:
Fixes #228

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -925,22 +925,22 @@ Here is an example of headers needed to sign a request.
 
 ~~~~~
   {
-    "(request-target)": "post /path",
+    "@request-target": "post /path",
     "content-length": 380,
     "date": "Mon, 08 Jul 2024 14:16:20 GMT",
-    "digest": "SHA-256=U7gNVUQiixe5BRbp4Tg0xCZMTcSWXXUZI2\\/xtHM40S0=",
+    "content-digest": "SHA-256=U7gNVUQiixe5BRbp4Tg0xCZMTcSWXXUZI2\\/xtHM40S0=",
     "host": "hostname.of.the.recipient",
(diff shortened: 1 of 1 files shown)


IETF-RFC.md
KrausMatthias, line 1: I think you missed some occurences of (request-target) and digest (somehow in github I can't comment on lines to far away from your changes).

@request-target is not a required part of the signature?

KrausMatthias (changes requested):

IETF-RFC.md
glpatcern: Right, now I went for a full search without just relying on my memory ;-)

glpatcern:
@KrausMatthias if `@request-target` is optional, we may just omit it from the example, just to keep it the simplest possible. Please have a look at it now

IETF-RFC.md
KrausMatthias: LGTM now :)

I don't know if droping @request-target from required signature parameters was intended though, just noticed it is not listed at https://github.com/cs3org/OCM-API/pull/231/files#diff-371e2d091f4292c62681d09a829854e2834ac93a195d6e17727927f9f0bf1344L1004 right now

KrausMatthias (approved):

IETF-RFC.md
glpatcern: Yep, as commented, if @request-target is optional, we may just omit it from the example to keep it simpler. After all this is an example, not a prescriptive part of the spec :)

mickenordin (approved):

PR #233, opened 2025-08-04
fix: description based on IETF I-D

MahdiBaghbani:
During a review with @KrausMatthias, we found out the OpenApi spec is a bit outdates, I'll go through the changes and try to sync it.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -548,7 +548,7 @@ components:
         resourceType:
           type: string
           description: |
-            Resource type (file, folder, calendar, contact, ...)
+            Resource type (file, calendar, contact, ...)
           example: file
         expiration:
           type: integer
@@ -771,7 +771,7 @@ components:
         resourceType:
           type: string
           description: |
-            Resource type (file, folder, user, calendar, contact, ...)
+            Resource type (file, user, calendar, contact, ...)
         providerId:
           type: string
           description: |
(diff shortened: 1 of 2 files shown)


spec.yaml
```diff
@@ -348,15 +348,18 @@ components:
...
+                  contain more than one resource type object per given `name`
                 example: file
               shareTypes:
                 type: array
                 description: |
-                  The supported recipient share types.
-                  Implementations MUST support `user` at a minimum.
+                  The supported recipient share types. 
+                  MUST contain `"user"` at a minimum, plus optionally 
+                  `"group"` and `"federation"`. Example: `["user"]`
```
```suggestion
                  `"group"` and `"federation"`.
```
(the example is given below)

spec.yaml
```diff
@@ -348,15 +348,18 @@ components:
               name:
                 type: string
                 description: >
-                  A supported resource type (file, folder, calendar, contact,
-                  ...).
-                  Implementations MUST support `file` at a minimum.
+                  A supported resource type (file, folder, calendar, contact, ...).
```
glpatcern: The formatting was adjusted to fit 72 chars per line per IETF constraint. We could actually add a linting action to check for that.
```suggestion
                  A supported resource type (file, calendar, contact, ...). 
```

glpatcern (changes requested):
Well spotted, yet I have a couple of small things to fix

glpatcern (approved):

Issue #229, opened 2025-08-01
How to get the Sending Server from Notifications without request signature


KrausMatthias:
Hi, 

I'm currently trying to implement the 1.2.0 spec in [OpenCloudMesh-rs](https://codeberg.org/OpenGeoMesh/OpenCloudMesh-rs).

As far as I understand the ```providerId``` identifying a share is only unique per provider (Sending Server), so to uniquely identify a share I need ```providerId``` and the Sending Server.

If there is a request signature, I could get the Sending Server via the `keyId`, but as the request signature is optional I don't necessarily have that and I don't see how I would get the Sending Server then, e.g. from a `SHARE_UNSHARED` notification?

mickenordin:
You should get the sending server by parsing the sender field, that means splitting the string on `@` and taking _the last_ part. For instance if I send you a share, the sender will be: `kano@sunet.se@sunet.drive.sunet.se` in that case, the sending server is `sunet.drive.sunet.se`.

mickenordin:
The process is discribed here: https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md#process

And by the way, very cool that you are doing a rust implementation of OCM. You are welcome to join the matrix channel if you haven't already: https://matrix.to/#/#cs3org_OCM:gitter.im if you like to chat or ask questions.

KrausMatthias:
Thank you! As far as I can tell the `sender` is only contained in the [Share Creation Notification](https://github.com/cs3org/OCM-API/blob/develop/spec.yaml#L474) to `/shares`, but not in the [Notifications](https://github.com/cs3org/OCM-API/blob/develop/spec.yaml#L751) to `/notifications` ?

mickenordin:
> Thank you! As far as I can tell the `sender` is only contained in the [Share Creation Notification](https://github.com/cs3org/OCM-API/blob/develop/spec.yaml#L474) to `/shares`, but not in the [Notifications](https://github.com/cs3org/OCM-API/blob/develop/spec.yaml#L751) to `/notifications` ?

I see, yes, I think you are right. This looks like a spec bug, we reference Share Creation Notification for providerId, but we don't have sender in other notifications. We should have that.

glpatcern:
Closing this as the "fix" (use httpsig) was merged. We know we have a broader work to do with notifications, which goes beyond this specific issue.

PR #234, opened 2025-08-04
Notifications SHOULD use httpsig for sender authentification

KrausMatthias:
Alternative proposal for #229 which avoids adding additional required fields (#230)

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -805,6 +805,13 @@ of trusted users, along with any related share. The recipient MAY
 reciprocally remove that provider from the list of trusted users, along
 with any related share.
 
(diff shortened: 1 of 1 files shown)


glpatcern:
I like this proposal, which also goes in the direction of "securing" all endpoints.

Also, to overcome backwards compatibility issues, we have introduced the concept of `requirements` to basically signal the other party that a recommended behavior such as `SHOULD use httpsig` becomes a mandatory requirement. This would fit very well.

mickenordin (approved):
I think this is good, let's do it :)

glpatcern (approved):

PR #245, opened 2025-08-12
Cosmetic change + CI job to validate line length

glpatcern:
This PR introduces a double space after each full-stop as per IETF convention and as observed in existing RFCs.

Furthermore, we add a CI job to check that line lengths do not exceed 72 characters. Some more paragraph changes were done also in the appendix examples to pass the job.


mickenordin (approved):
LGTM

PR #246, opened 2025-08-12
Add introduction

mickenordin:


glpatcern (changes requested):
Micke, could you please format it to 72 chars?

Otherwise, reading it next to the abstract I have the impression that we could blend the two and avoid repetitions, what do you think?

mickenordin:
What do you think about this?

glpatcern:
In fact I realize I was vague in my comment: the thing is looking at the IETF guidelines you posted earlier, we need both the Abstract and the Introduction.

The Abstract could then be just a couple of sentences to capture the scope of OCM, including the Alice and Bob example, whereas the mention of 2015-2016 fits the Introduction. Do you want to give it a try?

mickenordin:
Good point, here is my attempt :)

MahdiBaghbani (approved):
This looks good! Thanks Micke

glpatcern (approved):
Good job!

PR #248, opened 2025-08-19
Proper url spec for ocm server url

MahdiBaghbani:
I'm intentionally not updating the Discovery endpoint (and any related fields) in this PR. This needs a brief design discussion first. For reference:
```
* REQUIRED: endPoint (string) - The URI of the OCM API available at
this endpoint.  Example: `"https://my-cloud-storage.org/ocm"`
```

Once we reach consensus, I'll submit a follow up patch with the corresponding edits.

This PR is a result of discussion in #244

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -1117,8 +1117,7 @@ format:
     * REQUIRED: `url` - an absolute URL identifying the 
     OCM Server. It MUST:
       * include scheme `http://` or `https://`
-      * include host 
-      (either a FQDN or IP address, with an optional PORT)
+      * include host (either a FQDN or an IP address)
       * MAY include a non-default port
       * MUST NOT include a base path (e.g., `/ocm`)
       * MUST NOT include userinfo, query, or fragment
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -1116,7 +1116,7 @@ format:
   of OCM Servers with the following string fields:
     * REQUIRED: `url` - an absolute URL identifying the 
     OCM Server. It MUST:
-      * include scheme `http://` or `https://`
+      * include scheme: either `https://` or (for testing purposes) `http://`
       * include host (either a FQDN or an IP address)
       * MAY include a non-default port
       * MUST NOT include a base path (e.g., `/ocm`)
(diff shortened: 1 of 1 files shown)


mickenordin (approved):
Great 👍

IETF-RFC.md
```diff
@@ -1114,20 +1114,31 @@ format:
   Servers exposed by the Directory Service
   * REQUIRED: `servers` - a JSON array of objects to describe the list
   of OCM Servers with the following string fields:
-    * REQUIRED: `url` - the OCM Server's FQDN
+    * REQUIRED: `url` - an absolute URL identifying the 
+    OCM Server. It MUST:
+      * include scheme `http://` or `https://`
+      * include host 
+      (either a FQDN or IP address, with an optional PORT)
```
```suggestion
      * include host (either a FQDN or an IP address)
```

IETF-RFC.md
```diff
@@ -1114,20 +1114,31 @@ format:
   Servers exposed by the Directory Service
   * REQUIRED: `servers` - a JSON array of objects to describe the list
   of OCM Servers with the following string fields:
-    * REQUIRED: `url` - the OCM Server's FQDN
+    * REQUIRED: `url` - an absolute URL identifying the 
+    OCM Server. It MUST:
+      * include scheme `http://` or `https://`
```
```suggestion
      * include scheme: either `https://` or (for testing purposes) `http://`
```

glpatcern (changes requested):
Looks good! Just a couple of changes

KrausMatthias:
Maybe we should add something in Discovery on how to handle https vs http. Default to https only, rejecting any federated urls with http or OCM Servers only discoverable via http. Allow an Opt-In for http downgrades/fallback?

One might also leave that up to implementations, just something I've been thinking about.

glpatcern:
> Maybe we should add something in Discovery on how to handle https vs http. Default to https only, rejecting any federated urls 

Yes, that's what Mahdi put in the subject of this PR, leaving Discovery as a separate point.

My take here would be to provide a recommendation (the obvious one...) and leave it to the implementations. De facto the only room for `http` is for testing purposes, and even there we have certificates and https in the OCM Test Suite, Paired with the Security considerations where we recommend httpsig, it does not leave much choice,

glpatcern (approved):

PR #249, opened 2025-08-20
Further fixes and clarifications on the share fields

glpatcern:
Following #244, I realized that the text describing the share fields had some residual references from past edits, which make no sense any longer. This is to fix them.


mickenordin (approved):
:+1:

mickenordin:
Can you rebase on current develop?

MahdiBaghbani (approved):
Richard has changed the `resource` to `Resource` in the I-D, but in the `spec.yaml` we are still using `resource`.

I've merged the current develop into this branch with small modification and change the one-off `Resource` to `resource` for consistency

PR #255, opened 2025-08-20
add: CI for GitHub pages

MahdiBaghbani:
Based on Michiels implementation of `docs.html` we don't even need to build GitHub pages on `spec.yaml` changes since the script inside the page would fetch latest spec.

```html
<script>
      const params = new URLSearchParams(window.location.search);
      const user = params.get('user');
      const repo = params.get('repo');
      const branch = params.get('branch');
      let changed = false;
      if (user === null) {
        params.set('user', 'cs3org');
        changed = true;
      }
      if (repo === null) {
        params.set('repo', 'OCM-API');
        changed = true;
      }
      if (branch === null) {
        params.set('branch', 'develop');
        changed = true;
      }
      if (changed) {
        params.sort();
        window.location = `?${params.toString()}`;
      }
      document.querySelector('redoc').setAttribute('spec-url', `https://raw.githubusercontent.com/${user}/${repo}/${branch}/spec.yaml`);
    </script>
```

### DIFF
--- a/.github/workflows/github-pages.yml
+++ b/.github/workflows/github-pages.yml
@@ -1,0 +1,42 @@
+name: Deploy GitHub Pages
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Build site
+        run: |
+          rm -rf dist
+          mkdir -p dist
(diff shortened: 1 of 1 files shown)


glpatcern (approved):
OK, so the idea is to have the CI run on demand, but most of the time we won't even need to run it?

MahdiBaghbani:
Yes, we just need to run it once and then play with URL query

PR #257, opened 2025-08-20
CI backward compatibility

MahdiBaghbani:
The URL `https://cs3org.github.io/OCM-API/docs.html` is referenced in lot of places, so it should remain same


glpatcern (approved):
Right, and in fact the `README.md` gets currently rendered as landing page, which is good

MahdiBaghbani:
If want to keep it I need to change a bit

PR #259, opened 2025-08-20
Fix references and format

mickenordin:
This fixes the reference and boiler plate issues, Unfortunately it is also a bit messy since a lot of other formatting happened here as well :grimacing:


MahdiBaghbani:
We are all having the same problem, I also need to open yet another PR :smile:

glpatcern (approved):
Good!

Issue #237, opened 2025-08-06
Canonical format for fields in API Specification based on RFCs


MahdiBaghbani:
This issue has been raised on the test suite before between @glpatcern and I:
https://github.com/cs3org/ocm-test-suite/issues/157

Recent suggestion from @KrausMatthias:
> What do you think about defining `shareWith` etc. to be acct URIs? https://www.rfc-editor.org/rfc/rfc7565 From what I've understood that would be backwards compatible (when omitting the acct: scheme) and would make the requirement of the `user@host` format more explicit than just with the current example?

from @mickenordin: 
> What is the benefit in your opinion? I don't see how we can use the acct: scheme for ocm-address' since we want them to be backwards compatible and what is left is the normal ocm-address as defined to day?

mickenordin:
I was under the impression that user@sub.domain.tld was used everywhere, but obvious I am mistaken.

Nextcloud has shown the ocm address to users in this format since at least version 19, but you are saying that it is not entered that way in the various json payloads?

MahdiBaghbani:
Based on a detailed logging in https://github.com/cs3org/ocm-test-suite/issues/156, the JSON payloads are definitely different.

Also, Nextcloud didn't strictly follow the correct names is the JSON payload and has been fixed by Michiel and you via the
https://github.com/nextcloud/server/pull/47199
and
https://github.com/nextcloud/server/pull/50069

MahdiBaghbani:
Nextcloud still maintains additional properties in the JSON, `sharedBy`, `sharedByDisplayName` for backward compatibility.

I need to check the sources again and see how Nextcloud handles incoming shares, does it require the `sharedBy` or not.

KrausMatthias:
I just thought the acct URI would match very closely the semantics of the OCM Address

> OCM Address - a string of the form <Receiving Party's identifier>@<fqdn> which can be used to *uniquely identify a user or group "at" an OCM Server.* <Receiving Party's identifier> is an opaque string, unique at the server. <fqdn> is the Fully Qualified Domain Name by which the server is identified. This MUST be the domain at which the /.well-known/ocm endpoint of that server is hosted.
> https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md?plain=1#L103
---
> This document defines the 'acct' Uniform Resource Identifier (URI)
   scheme as a way to *identify a user's account at a service provider,*
   irrespective of the particular protocols that can be used to interact
   with the account.

Pros
[...]

PR #268, opened 2025-08-22
Proposal for the OCM Address format

KrausMatthias:
Following discussion in #237 and #244

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -514,9 +514,9 @@ Step 4: If not, try a HTTP GET with `https://<fqdn>/ocm-provider` as
 the URL instead.
 Step 5: If that results in a valid HTTP response with a valid JSON
 response body within reasonable time, go to step 7.
-Step 6: If not, fail. Implementations MAY fallback to `http` instead
-`https` in testing setups and retry steps 2-5 using `http`. This SHOULD
-only be allowed as Opt-In.
+Step 6: If not, fail. Implementations MAY fallback to HTTP instead
+of HTTPS in testing setups and retry steps 2-5, in particular when
+an optional port is given in the address.
 Step 7: The JSON response body is the data that was discovered.
 
 ## Fields
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -516,7 +516,9 @@ Step 4: If not, try a HTTP GET with `https://<fqdn>/ocm-provider` as
 the URL instead.
 Step 5: If that results in a valid HTTP response with a valid JSON
 response body within reasonable time, go to step 7.
-Step 6: If not, fail.
+Step 6: If not, fail. Implementations MAY fallback to `http` instead
+`https` in testing setups and retry steps 2-5 using `http`. This SHOULD
+only be allowed as Opt-In.
 Step 7: The JSON response body is the data that was discovered.
 
 ## Fields
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -139,14 +139,31 @@ related concepts from OAuth [RFC6749] and elsewhere:
   OCM API Discovery.
 * __Discoverable Server__ - A server that tries to supply information in
   OCM API Discovery.
-* __OCM Address__ - A string of the form
-  `<Receiving Party's identifier>@<fqdn>` which can be used to uniquely
-  identify a user or group "at" an OCM Server and MAY be referred to as
-  Federated Cloud ID.
-  `<Receiving Party's identifier>` is an opaque string, unique at the
-  server.  `<fqdn>` is the Fully Qualified Domain Name by which the
-  server is identified.  This MUST be the domain at which the
-  `/.well-known/ocm` endpoint of that server is hosted.
+* __OCM Address__ - identifies a user or group "at" an OCM Server.
+The OCM Address contains a server specific Party identifier, a host
+locating the OCM Server and an optional port. The OCM Address is not a
+URI as it does not have scheme and the identifier may contain reserved
+characters.
+
+        ocm-address = identifier "@" host [ ":" port]
+
+    The identifier is an opaque, case-sensitive utf-8 string. It is 
+    separated from the host by the last "@" in the OCM Address.
+    host is an IP literal encapsulated within square brackets, an IPv4 
+    address in dotted decimal form, or a registered name as described in
+    [RFC3986].
+
+        host = IP-literal / IPv4address / reg-name
+
+    The optional port subcomponent can be used to specify a port to use
+    for discovery. Implementers SHOULD omit the port component and its 
+    ":" delimiter if its value would be the same as the default ports 
+    "443" or "80" of HTTPS or HTTP.
+
+    The OCM Server MUST be discoverable at the given host and optional 
+    port via the Well-Known [RFC8615] path `/.well-known/ocm`. The OCM 
+    Address MUST NOT contain a path.
+
 * __OCM Notification__ - A message from the Receiving Server to the
   Sending Server or vice versa, using the OCM Notifications endpoint.
 * __Invite Message__ - Out-of-band message used to establish contact
@@ -1000,6 +1017,13 @@ Key Words](https://datatracker.ietf.org/html/rfc8174)", May 2017.
 
 [RFC9421] Backman, A., Richer, J. and Sporny, M. "[HTTP Message
 Signatures](https://tools.ietf.org/html/rfc9421)", February 2024.
+
+[RFC3986] Berners-Lee, T., Fielding, R. and Masinter, L. 
+"[Uniform Resource Identifier (URI): Generic Syntax
+](https://datatracker.ietf.org/doc/html/rfc3986)", January 2005
+
+[RFC8615] Nottingham, M. "[Well-Known Uniform Resource Identifiers 
+(URIs)](https://datatracker.ietf.org/doc/html/rfc8615)", May 2019
 
 
 ## Informative References
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -147,7 +147,7 @@ characters.
 
         ocm-address = identifier "@" host [ ":" port]
 
-    The identifier is an opaque, case-sensitive utf-8 string. It is 
+    The identifier is an opaque, case-sensitive UTF-8 string. It is 
     separated from the host by the last "@" in the OCM Address. It is
     possible to have multiple @-signs in a OCM-address, e.g. when an 
     email address is the local part of the address like 
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -148,7 +148,11 @@ characters.
         ocm-address = identifier "@" host [ ":" port]
 
     The identifier is an opaque, case-sensitive utf-8 string. It is 
-    separated from the host by the last "@" in the OCM Address.
+    separated from the host by the last "@" in the OCM Address. It is
+    possible to have multiple @-signs in a OCM-address, e.g. when an 
+    email address is the local part of the address like 
+    `nomen.nescio@example.org@ocm.example.org`.
+
     host is an IP literal encapsulated within square brackets, an IPv4 
     address in dotted decimal form, or a registered name as described in
     [RFC3986].
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -160,9 +160,7 @@ characters.
         host = IP-literal / IPv4address / reg-name
 
     The optional port subcomponent can be used to specify a port to use
-    for discovery. Implementers SHOULD omit the port component and its 
-    ":" delimiter if its value would be the same as the default ports 
-    "443" or "80" of HTTPS or HTTP.
+    for discovery (see Discovery Process).
 
     The OCM Server MUST be discoverable at the given host and optional 
     port via the Well-Known [RFC8615] path `/.well-known/ocm`. The OCM 


KrausMatthias:
BTW what is the __Sharing User__ compared to the __Sending Party__ ?

Should we replace the __FQDN__ by some __OCM Server Address__ consisting of `host[":" port]`?

IETF-RFC.md
```diff
@@ -139,14 +139,31 @@ related concepts from OAuth [RFC6749] and elsewhere:
...
+* __OCM Address__ - identifies a user or group "at" an OCM Server.
+The OCM Address contains a server specific Party identifier, a host
+locating the OCM Server and an optional port. The OCM Address is not a
+URI as it does not have scheme and the identifier may contain reserved
+characters.
+
+        ocm-address = identifier "@" host [ ":" port]
+
+    The identifier is an opaque, case-sensitive utf-8 string. It is 
+    separated from the host by the last "@" in the OCM Address.
```
mickenordin: The fact that it is the `last` @-sign is important, since it is common for example for an email-address to be the local part of the identifier, maybe we should say that explicitly, e.g.:

```
It is separated from the host by the last "@" in the OCM Address. It is possible, and common, to have multiple @-signs in a OCM-address, e.g. when an email address is the local part of the address:
  nomen.nescio@example.org@ocm.example.org
```

KrausMatthias: In URIs that would be avoided by requiring @, :, / etc. to be percent-encoded

mickenordin (approved):
I think this is good!

IETF-RFC.md
```diff
@@ -139,14 +139,35 @@ related concepts from OAuth [RFC6749] and elsewhere:
...
-  `/.well-known/ocm` endpoint of that server is hosted.
+* __OCM Address__ - identifies a user or group "at" an OCM Server.
+The OCM Address contains a server specific Party identifier, a host
+locating the OCM Server and an optional port. The OCM Address is not a
+URI as it does not have scheme and the identifier may contain reserved
+characters.
+
+        ocm-address = identifier "@" host [ ":" port]
+
+    The identifier is an opaque, case-sensitive utf-8 string. It is
```
```suggestion
    The identifier is an opaque, case-sensitive UTF-8 string. It is 
```

IETF-RFC.md
```diff
@@ -139,14 +139,35 @@ related concepts from OAuth [RFC6749] and elsewhere:
...
+    host is an IP literal encapsulated within square brackets, an IPv4 
+    address in dotted decimal form, or a registered name as described in
+    [RFC3986].
+
+        host = IP-literal / IPv4address / reg-name
+
+    The optional port subcomponent can be used to specify a port to use
+    for discovery. Implementers SHOULD omit the port component and its 
+    ":" delimiter if its value would be the same as the default ports 
+    "443" or "80" of HTTPS or HTTP.
```
glpatcern: Thanks for making this explicit now. IMHO we should be a bit more restrictive, also following the current implementations:
```suggestion
    for discovery. If no port is specified, HTTPS on the default port
    443 MUST be assumed. If a port is specified, implementers MAY
    use HTTPS or HTTP, in order to ease designing unit tests.
```

glpatcern (changes requested):
Looks good, but at this point I propose to make the following change:

IETF-RFC.md
KrausMatthias: Maybe we should put that in the Discovery Process?

glpatcern: That's definitely an option (it was too far to edit that section from this change), though a repetition might not hurt. You could try and take this paragraph down to the Discovery and put here a forward reference - let's see how it reads.

MahdiBaghbani:
Based on the discussion in #270 
Should we standardize concrete address forms (or consider it) for each entity type (e.g. `user@fqdn`, `group:slug@fqdn`, `federation:tag`) to eliminate ambiguity?

I believe it at least heps with parsing ambiguity and validating the share notification

I can think of an example based on this PR:
```
ocm-address = identifier "@" server

user-addr: identifier = user name or user id, server = fqdn [ ":" port] or host [ ":" port]
group-addr: identifier = "group:" (group-id / group-slug) "@" server = fqdn [ ":" port] or host [ ":" port]
federation-addr: "federation:" fqdn [ ":" port] or host [ ":" port] or fed-tag or fed-id
```
in case of `fqdn [ ":" port] or host [ ":" port]` and `federation` type, the share is emant for all people at the `fqdn`
in case of `fed-tag` or `fed-id` (wedont have `fed-id` in spec yet), the share is meant for all people in all servers in the said federations

This is better to go on a new PR where we can talk more about it's design but it worth to mention it here.

IETF-RFC.md
```diff
@@ -139,14 +139,35 @@ related concepts from OAuth [RFC6749] and elsewhere:
...
+
+        ocm-address = identifier "@" host [ ":" port]
+
+    The identifier is an opaque, case-sensitive UTF-8 string. It is 
+    separated from the host by the last "@" in the OCM Address. It is
+    possible to have multiple @-signs in a OCM-address, e.g. when an 
+    email address is the local part of the address like 
+    `nomen.nescio@example.org@ocm.example.org`.
+
+    host is an IP literal encapsulated within square brackets, an IPv4
```
mickenordin, line 156: Maybe we should have examples of each of the formats? I think that the version with en ip address encapsulated in square brackets might break current implementations, we should test that to make sure.

mickenordin:
I also just noticed something "fun", this is the ocm address as presented by owncloud 10 from surfbox, where I am logged in with my eduid.nl account:

` Your Federated Cloud ID: def1d354-a6b8-43a5-ab69-70bea254361c@eduid.nl@surfdrive.surf.nl/files`

notice the path at the end of the hostname. Shoudl we infact allow that? That will make instances installed in subpaths work.

KrausMatthias:
> I also just noticed something "fun", this is the ocm address as presented by owncloud 10 from surfbox, where I am logged in with my eduid.nl account:
> 
> ` Your Federated Cloud ID: def1d354-a6b8-43a5-ab69-70bea254361c@eduid.nl@surfdrive.surf.nl/files`
> 
> notice the path at the end of the hostname. Shoudl we infact allow that? That will make instances installed in subpaths work.

Discovery via the .well-known/ocm path is not possible in a subfolder, how should discovery behave in the case of subfolders?

IETF-RFC.md
KrausMatthias: I guess square brackets are a prerequisite for IPv6 support, so it should be supported by all IPv6 capable implementations.
@MahdiBaghbani  Maybe it would be possible to add a testcase in ocm-test-suite?

MahdiBaghbani: @KrausMatthias yes, we would have a matrix for permutations of send/receive/fqdn/ip/port across multiple platforms :smile:

It's going to be huge matrix, but I think the result would give us a great insight of how OCM is being implemented for future spec discussions

KrausMatthias:
> Based on the discussion in #270 Should we standardize concrete address forms (or consider it) for each entity type (e.g. `user@fqdn`, `group:slug@fqdn`, `federation:tag`) to eliminate ambiguity?
> 
> I believe it at least heps with parsing ambiguity and validating the share notification
> 
> I can think of an example based on this PR:
> 
> ```
> ocm-address = identifier "@" server
> 
> user-addr: identifier = user name or user id, server = fqdn [ ":" port] or host [ ":" port]
> group-addr: identifier = "group:" (group-id / group-slug) "@" server = fqdn [ ":" port] or host [ ":" port]
> federation-addr: "federation:" fqdn [ ":" port] or host [ ":" port] or fed-tag or fed-id
> ```
> 
> in case of `fqdn [ ":" port] or host [ ":" port]` and `federation` type, the share is emant for all people at the `fqdn` in case of `fed-tag` or `fed-id` (wedont have `fed-id` in spec yet), the share is meant for all people in all servers in the said federations
> 
> This is better to go on a new PR where we can talk more about it's design but it worth to mention it here.

I agree there is a gap where users or implementations need to know if a certain OCM Address identifies a user, group or federation to be able to fill the `shareType` field in the `NewShare` message.

If we make that part of the OCM Address we should drop the `shareType` field in `NewShare`.

"group:" or "federation:" prefixes might already be contained in the existing opaque strings so that would be a breaking change.

If we do that, I would be strongly in favor of making OCM Addresses a URI, with "group" and "federation" as URI scheme instead of defining something which looks like an URI but isn't.

I'm wondering if we can avoid needing the SendingServer to know it the recipient is a user, group or federation if we leave that distinction completely to the ReceivingServer (requiring a shared namespace on the ReceivingServer where a Group can't have the same name as a user)

IETF-RFC.md
MahdiBaghbani: Last time I worked with IPv6 in docker it wasn't great, maybe that would become a limitation in testing, I need to update myself on current status of IPv6 in docker networking

mickenordin:
> > I also just noticed something "fun", this is the ocm address as presented by owncloud 10 from surfbox, where I am logged in with my eduid.nl account:
> > ` Your Federated Cloud ID: def1d354-a6b8-43a5-ab69-70bea254361c@eduid.nl@surfdrive.surf.nl/files`
> > notice the path at the end of the hostname. Shoudl we infact allow that? That will make instances installed in subpaths work.
> 
> Discovery via the .well-known/ocm path is not possible in a subfolder, how should discovery behave in the case of subfolders?

I suppose, what we should do in case we receive such an OCM-address is to try https://surfdrive.surf.nl/.well-known/ocm and then fall back to the legacy endpoint at https://surfdrive.surf.nl/files/ocm-provider/ (which in fact, is what current implementations do as the first action)

MahdiBaghbani:
> If we do that, I would be strongly in favor of making OCM Addresses a URI, with "group" and "federation" as URI scheme instead of defining something which looks like an URI but isn't.

I agree, the blocker for making OCM addresses RFC7565 is the breaking change in already established user shares.

`federation` and `group` share are kind of new territory so we might be able to use URI in here?
@glpatcern and @mickenordin are better informed in this area, the current usage and implementation of it.

mickenordin:
> I'm wondering if we can avoid needing the SendingServer to know it the recipient is a user, group or federation if we leave that distinction completely to the ReceivingServer (requiring a shared namespace on the ReceivingServer where a Group can't have the same name as a user)

I would like to avoid having a centralized registry of federations if possible, because that makes it less of a federation and more of a distributed system, the sending server could of course be made to provide additional information about the federated share it wants to set up, and act as a proxy, but again, that just means sending multiple user or group shares to a multitude of servers. We should investigate what Nextcloud does with the federation share type before proceeding more with this line of inquery I think.

MahdiBaghbani:
Based on Nextcloud sources [create share](https://github.com/nextcloud/server/blob/32f5f6e08ebaccddf691806f2da812dd1f0416df/apps/federatedfilesharing/lib/FederatedShareProvider.php#L82-L232) and [notification](https://github.com/nextcloud/server/blob/32f5f6e08ebaccddf691806f2da812dd1f0416df/apps/federatedfilesharing/lib/Notifications.php#L36-L95)

They use 
```php
* send server-to-server share to remote server
*
* @param string $token
* @param string $shareWith
* @param string $name
* @param string $remoteId
* @param string $owner
* @param string $ownerFederatedId
* @param string $sharedBy
* @param string $sharedByFederatedId
* @param int $shareType (can be a remote user or group share)
 ```
 
In the notification it would change them to correct identifies for OCM in [here](https://github.com/nextcloud/server/blob/32f5f6e08ebaccddf691806f2da812dd1f0416df/apps/federatedfilesharing/lib/Notifications.php#L361-L377)

which finally sends the share in [here](https://github.com/nextcloud/server/blob/32f5f6e08ebaccddf691806f2da812dd1f0416df/lib/private/Federation/CloudFederationProviderManager.php#L106-L130)

The share type is either `user` or `group`, defined [here](https://github.com/nextcloud/server/blob/32f5f6e08ebaccddf691806f2da812dd1f0416df/lib/private/Federation/CloudFederationShare.php#L194-L200)

And this is for the [remote group](https://github.com/nextcloud/server/blob/32f5f6e08ebaccddf691806f2da812dd1f0416df/lib/private/Collaboration/Collaborators/RemoteGroupPlugin.php#L17-L73), see Nextcloud doesn't use `federation` but `group` for the shares
 
 I have a feeling I'm not searching the right places

File: IETF-RFC.md (hunk 3 above)
Comment (KrausMatthias) [reply IETF-RFC.md]: See f38d6af

Comment (glpatcern) [reply IETF-RFC.md]: OK very good. Then I'd still edit this as follows:
```suggestion
    for discovery (see Discovery Process).
```

IETF-RFC.md
```diff
@@ -495,7 +516,9 @@ Step 4: If not, try a HTTP GET with `https://<fqdn>/ocm-provider` as
 the URL instead.
 Step 5: If that results in a valid HTTP response with a valid JSON
 response body within reasonable time, go to step 7.
-Step 6: If not, fail.
+Step 6: If not, fail. Implementations MAY fallback to `http` instead
+`https` in testing setups and retry steps 2-5 using `http`. This SHOULD
+only be allowed as Opt-In.
```
```suggestion
Step 6: If not, fail. Implementations MAY fallback to HTTP instead
of HTTPS in testing setups and retry steps 2-5, in particular when
an optional port is given in the address.
```
I don't think we buy much in stating the Opt-in, rather the issue is this is for implementations to do in the context of testing, using custom ports as they wish.

IETF-RFC.md
KrausMatthias: Btw. I've implemented the suggested OCM Address Format in https://codeberg.org/OpenGeoMesh/OpenCloudMesh-rs/commit/488bf3a7fb802a2ffcc94d11d216934265b98219#diff-9f3595a4e934e00060a258d7c0d4d6daa7b7adcf

glpatcern (approved):
I think this changeset can go ahead as is. The current discussion about the semantic of group/federated sharing will surely continue in #270 but it's beyond this PR.

MahdiBaghbani (approved):
LGTM! Thanks Matthias

Issue #270, opened 2025-08-22
Improve specification for sharing to `group` or `federation`


glpatcern:
The current [Share Creation Notification](https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md#share-creation-notification) includes a `shareType` field that `SHOULD` (actually, `MUST`) be one of `user`, `group`, or `federation`, where the latter cases correspond to:
* `group`: a group of users at the Receiving Server~~s~~ (should be a single `Receiving Server`)
* `federation`: a group of users that is spread out over various servers, including at least one user at the Receiving Server

The issue is that there's no provision about how to advertise or inform parties about the identifiers of such groups, in particular when it comes to a "federation" that spans multiple servers.

This is a topic that indirectly came up as part of the current build-up of the EOSC Nodes, given that some kind of global/federated AAI is sought after, and this is where such "OCM-sharing to a group or federation" would fit. In that context, a Directory Service may also host a sort of address book of federations. But in general, an OCM Server may also provide the same functionality (IIUC this is what Nextcloud names Circles).

I think we should expand the OCM specification in order to contemplate such options. Any thoughts @sunetfreitag @mickenordin ?

KrausMatthias:
One alternative approach might be a OCM Server acting as proxy. The Proxy would receive the federated share, and forward it to the actual recipients, posing as sending server. That way the members of a federated group could stay private.

We also need to make sure to prevent loops of federated shares or forbid nested federated groups entirely

sunetfreitag:
I went through the current document and currently have more questions than answers. I will work on the answers during the next few days. Here are my thoughts:

The current definitions and relations between 'federation' shareType, groups, federated groups, directory service and the share creation notification are ambiguous and require more specification and definition, especially their relationship to ocm address. There are currently eight references to federation in current specification.

**General references to federation:**

- Abstract: "Open Cloud Mesh (OCM) is a server federation protocol"
- Invite Acceptance Request Details mentions federation in the context of the WAYF Page: "[...] if the Invite Sender is part of a federation of associated OCM Servers."
- Trust security considerations: "(...) the Directory Service feature can be used to establish a trusted federation"

**Technical references to federation:**

- OCM API Discovery: shareTypes as part of the resourceTypes field: "The supported recipient share types. MUST contain "user" at a minimum, plus optionally "group" and "federation""
- Share Creation Notification field: "shareWith" is the OCM Address of the user, group or federation (REQUIRED)
[...]

PR #272, opened 2025-08-25
fix: group is meant for the users on a singular server

MahdiBaghbani:
PR for fixing a typo pointed out this discussion: https://github.com/cs3org/OCM-API/issues/270#issuecomment-3219699828


IETF-RFC.md
```diff
@@ -672,7 +672,7 @@ To create a Share, the Sending Server SHOULD make a HTTP POST request
   SHOULD have a value of "user", "group", or "federation", to
```
glpatcern, line 672:
```suggestion
  MUST have a value of "user", "group", or "federation", to
```

glpatcern (approved):

PR #273, opened 2025-09-02
Fix: explicitly state the uniqueness of providerIds

### LINKED ISSUE #271: ProviderId Clarification
Reading through the #194 and #196:

Clarification on `providerId` semantics

The [text](https://github.com/cs3org/OCM-API/blob/dced0610f9750982bc2035d12e457f883eded77d/IETF-RFC.md?plain=1#L651-L656) says:

> This is unique per provider.
> If the same resource is shared twice, this `providerId` will not be repeated.

Read together, they're ambiguous. Two possible interpretations:

Resource scoped: `providerId` identifies a resource within the provider. If the same resource is shared multiple times, the same `providerId` should appear each time

Share scoped (per receiver): `providerId` identifies a share instance created by the provider. If the same resource is shared to two receivers, the `providerId`s differ. This aligns with Nextcloud/ownCloud, where share IDs are created sequentially in the share table.

glpatcern:
This fixes #271 with an explicit sentence about the uniqueness of `providerId` values in Share Creation Notification payloads.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -649,10 +649,9 @@ To create a Share, the Sending Server SHOULD make a HTTP POST request
   Example: "This is the Open API Specification file (in YAML
   format) of the Open Cloud Mesh API."
 * REQUIRED providerId (string)
-  Identifier to identify the Shared Resource at the provider
-  side.  This is unique per provider such that if the same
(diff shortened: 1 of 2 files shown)


MahdiBaghbani (approved):
Nice!

PR #278, opened 2025-10-01
Spell out how federated sharing should work

glpatcern:
I tried to include the options discussed in #270, ~~not including the Discovery Service though as IMHO that's a different service/function~~. Of course open for further expansions.

Edit: the Discovery Service is now explicitly mentioned, as well as the "OCM proxy" mode, while leaving the options to resolve members of groups fully open.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -535,7 +535,7 @@ contain the following information about its OCM API:
 * REQUIRED: resourceTypes (array) - A list of all resource types this
   server supports in both the Sending Server role and the Receiving
   Server role, with their access protocols.  Each item in this list
-  SHOULD itself be an object containing the following fields:
+  MUST itself be an object containing the following fields:
   - name (string) - A supported resource type (file, calendar,
     contact, ...).
     Implementations MUST offer support for at least one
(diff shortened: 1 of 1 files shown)


mickenordin (changes requested):
I think we need to spell out what a "federation" is as well.

An OCM server can be part of one or more federations.

A federation is a named list of OCM servers. An OCM server can know which federations it is a member of and which peers it has, either by static configuration (config file, database configuration) or by using one or more discovery services laid out in the appendix.

When a federated share is made on the sharing server, it targets one of the federations that the OCM server is part of, and the result should be that the server sends an individual share to each of the servers in the federation.

I don't think we should have any other central AAI server that acts as a proxy, apart from the discovery service already specified. I think it will complicate things without need, and will move us a way from the mesh/federation thinking to a centalized protocol which I don't think we actually need.

glpatcern:
Interesting observation. So far, the federation in the sharing to a `federation` gesture and the federation exposed by the Directory Service have been two completely independent concepts, as outlined in #270, though we use the same word.

I agree we should clarify what we intend with "federation", and whether to bind the two concepts, as @mickenordin proposes. I see the advantage of going for a more coherent specification, but also the disadvantage that existing (Nextcloud) implementations that offer the sharing to `federation` but are not using a Directory Service in the OCM sense become out of spec.

Concerning federated AAI services, I think we don't need anything specific in the OCM specs, but I expect they will exist in one form or another, so contemplating their existence may help.

glpatcern:
Following discussions, it turns out that Nextcloud does not implement the "sharing to `federation`" gesture with OCM - this feature remained in an unmerged PR.

I will then edit the PR to link the sharing to `federation` with the federations exposed by the Directory Service, opening the possibility to extend invites to groups, without fully specifying that for the time being. The goal here is to make the spec consistent with what we have.

mickenordin (approved):
I am ok with this wording, but I think we'll need to revisit it in the future when we actually implement it.

glpatcern:
> I am ok with this wording, but I think we'll need to revisit it in the future when we actually implement it.

Yep, I think so as well. Similarly to the Notifications part, where we have some `this is out of scope in this version` statements.

PR #280, opened 2025-10-02
Clarify re-sharing is not implicitly allowed

### LINKED ISSUE #212: Resharing semantic and constraints

glpatcern:
Pending the full specification of Notifications, this patch removes a vague reference to the possibility of re-sharing by a Receiving Party, and makes it explicit that this requires explicit consent.

The flow would require the Receiving Server to send a `REQUEST_RESHARE` Notification back to the Sender and look for a yet-to-be-specified response, with the re-share being executed by the Sender Server "on behalf" of the Receiving Party.

This partially fixes #212, though a full fix implies fully specifying the `REQUEST_RESHARE` Notification.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -223,8 +223,9 @@ Notification to the Receiving Server.
 
 After this, the Receiving Server MAY notify the Receiving Party and/or
 the Sending Server, and will act as an API client through which the
-Receiving Party can access the Resource.  After that, the Share MAY be
-updated, deleted, and/or reshared.
+Receiving Party can access the Resource.  The Receiving Party or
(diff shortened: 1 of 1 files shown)


mickenordin (approved):
👍

Issue #276, opened 2025-09-24
RFE: OCM Address Linking


sunetfreitag:
While an ocm address is a convenient way of identifying a sending and receiving party, users might be faced with a situation where they might be associated with multiple ocm addresses. In those cases, linking of ocm addresses might be beneficial, e.g., in case someone might want to share resources with "themselves". In case of (verified) account linking, resource sharing could be automated, without having to accept a share creation notification.

glpatcern:
@sunetfreitag could you please elaborate with an example?

Here follows my own understanding: the desired linking should happen within the EFSS at site `S` between e.g. a `lopresti` account associated with `giuseppe.lopresti@cern.ch` and registered in the `S` local accounting system, and the OCM address `lopresti@cernbox.cern.ch`, thanks to the fact that the latter is associated with the _same_ email address, following a successful OCM invite by some user belonging to `S`. So it's a linking between a "local" account and an OCM address. How would users have _multiple_ OCM addresses in a given system?

An extension of this concept would apply when a user A wants to create a share to an external user B. Multiple scenarios are possible:
* B is not known at all: the system could propose A to generate an OCM invite to B's email, and then create the OCM share upon receiving acceptance of the invite.
* B is only known as "local" user: a regular share is created; possibly a prompt is shown to A whether an OCM invite is to be sent to B?
* B is only known as OCM user: an OCM share is created. This is the only implemented behavior for all vendors.
* B is known to A's system both as "local" user AND as OCM user: both a regular share AND an OCM share should be created, as the two accounts are "linked".

I give credit to @labkode to having proposed the first scenario above some time ago, but I think it can be generalized as described.

Finally let me also say that while it's useful to add some recommendations in the OCM specs around the invites, ultimately the heavy lifting is entirely with the implementations, so I suggest to raise an issue at the vendor's GitHub tracker for the desired behavior.

mickenordin:
I think we should be weary of adding too much centralization into the protocol. Being able to link accounts across multiple instances means some kind of central infrastructure, this goes against the idea of federation, and takes us in to the realm where a lot of infrastructure is needed and it is hard to self host.

Already the discovery service is a step in this direction, even though it is not mandatory for invites, but only a convenience when available.

labkode:
For the discovery part, this may be worth revisiting: https://indico.cern.ch/event/1332413/contributions/5749456/attachments/2817731/4919651/ocm-dns.pdf

glpatcern:
[...]

PR #283, opened 2025-10-06
Introduce optional linking of a local and an OCM address

glpatcern:
Addresses #276 for the part concerning the possible link between a local identity of an external user and the corresponding OCM address, when the email addresses match.


mickenordin (approved):
Approve as per discussion on the call

PR #285, opened 2025-10-16
Clarify resourceType use in new shares

glpatcern:


mickenordin (approved):
Agreed

KrausMatthias:
Nextcloud rejected shares with resourceType 'folder' during my testing.

glpatcern:
> Nextcloud rejected shares with resourceType 'folder' during my testing.

That was my experience as well. Once Nextcloud supports the new 'multi-protocol' format (which is already being worked on by @mickenordin), it should handle `folder` shares as well.

KrausMatthias:
Would that mean that "folder" needs to be advertised as well in discovery in addition to "file" in discovery?
"folder" shares should probably only be sent to OCM Receivers which declare to support "folder" resources and instead be sent as "file"?

Existing implementations expect "file or folder" for "file" shares and will continue to send folder shares as "file". So after this change Receiving Servers would still need to check if the share is a folder for "file" shares?

glpatcern:
> Would that mean that "folder" needs to be advertised as well in discovery in addition to "file" in discovery? "folder" shares should probably only be sent to OCM Receivers which declare to support "folder" resources and instead be sent as "file"?

Expanding here some of the comments in the chat: in principle that would make sense, yes, and exposing `folder` in the discovery would implicitly signal that the system is capable of handling files and folders separately.

> Existing implementations expect "file or folder" for "file" shares and will continue to send folder shares as "file". So after this change Receiving Servers would still need to check if the share is a folder for "file" shares?

Indeed, the change is just a recommendation ("SHOULD") - Receiving Servers would still need to check every received share, or they could cache the discovery payload and skip the check in case `folder` is there.

The reality though is that the implementations that send `folder` are not advertising it in the discovery, whereas they should. To be followed up with the implementations, at least the Reva-based ones: @MahdiBaghbani this is another small change in the direction of making shares fully working.

PR #287, opened 2025-10-16
Preparing version 1.2.2

glpatcern:
As the IESG is at the final vote to establish a Working Group, I think it makes sense to tag a patch version now with all the refinements that went in so far.


MahdiBaghbani (approved):
LGTM

mickenordin (approved):
Let's go!

PR #291, opened 2025-10-29
Define what a Trusted Server is

mickenordin:


KrausMatthias:
How would we handle rotation of public keys?

mickenordin:
> How would we handle rotation of public keys?

I am not sure that we need to specify that, as we don't currently specify how to get the public key in the first place. But it is a good question in general, since we supply keys in the discovery and use http-sig in general. Maybe we should look in to how that works in http-sig?

glpatcern (approved):
This definition is good to me. Then, how to manage keys is definitely a topic to be looked after in the context of the IETF WG - as any other security-related issue.

PR #292, opened 2025-10-29
Make token endpoint configurable and align with OAUTH

mickenordin:
This patch makes the token endpoint configurable and anligns the request with the requirements with OIDC/OAUTH

* https://www.rfc-editor.org/rfc/rfc6749.html#section-4.1
* https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -645,7 +645,7 @@ contain the following information about its OCM API:
   to the more secure (and possibly required) invite flow.
   _ `"receive-code"` - to indicate that this OCM Server can receive a
   `code` as part of a Share Creation Notification, and exchange it
-  for a bearer token at the Sending Server's `/token` API endpoint.
+  for a bearer token at the Sending Server's tokenEndPoint.
   _ `"invite-wayf"` - to indicate that this OCM Server exposes a WAYF
   Page to facilitate the Invite flow.
 * OPTIONAL: criteria (array of string) - The criteria for accepting a
@@ -687,6 +687,11 @@ contain the following information about its OCM API:
   `"/index.php/apps/sciencemesh/accept"` is specified here then a WAYF
   Page SHOULD redirect the end-user to
 `/index.php/apps/sciencemesh/accept?token=zi5kooKu3ivohr9a&providerDomain=example.com`.
+* OPTIONAL: tokenEndPoint (string) - URL of the token endpoint where
+  the Sending Server can exchange a `code` for a bearer token.
+  Implementations that offer the `"receive-code"` capability MUST
+  provide this URL as well.
+  Example: `"https://my-cloud-storage.org/ocm/token"`.
 
 # Share Creation Notification
 
@@ -759,7 +764,7 @@ To create a Share, the Sending Server SHOULD make a HTTP POST request
   that the share does not expire.
 * OPTIONAL code (string)
   A nonce to be exchanged for a (potentially short-lived)
-  bearer token at the Sending Server's /token endpoint.
+  bearer token at the Sending Server's tokenEndPoint [RFC6749]
 * REQUIRED protocol (object)
   JSON object with specific options for each protocol.
   The supported protocols are: - `webdav`, to access the data -
@@ -961,9 +966,10 @@ is as follows:
     `resourceTypes[0].protocols.webdav` value is the
     `<sender-ocm-path>` to be used in step 3.
 2.  If `code` is not empty, the receiver SHOULD make a signed POST
-    request to the `/token` path inside the Sending Server's OCM API, to
+    request to the path in the Sending Server’s tokenEndPoint, to
     exchange the code for a short-lived bearer token, and then use that
-    bearer token to access the Resource.
+    bearer token to access the Resource (See the [Code Flow](
+    #code-flow) section).
 3.  If `protocol.name` = `webdav`, the receiver SHOULD inspect the
     `protocol.options` property.  If it contains a `sharedSecret`, as in
     the [legacy example](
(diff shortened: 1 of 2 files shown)


KrausMatthias:
I think the schema of the token response is missing?

mickenordin:
> I think the schema of the token response is missing?

Should be here, right: https://github.com/cs3org/OCM-API/blob/24b38584b2ef12cd049378a4daf3a5a15f9a8171/spec.yaml#L869

glpatcern:
This looks good, don't we also want to explicitly say that the token endpoint is expected to be "just OIDC"? Or what would make it different from a vanilla OIDC token endpoint?

mickenordin:
I think the difference is that there is supposed to be an authorize step before, that we skip, that is what will get you the refresh token.

IETF-RFC.md
```diff
@@ -687,6 +687,11 @@ contain the following information about its OCM API:
   `"/index.php/apps/sciencemesh/accept"` is specified here then a WAYF
   Page SHOULD redirect the end-user to
 `/index.php/apps/sciencemesh/accept?token=zi5kooKu3ivohr9a&providerDomain=example.com`.
+* OPTIONAL: tokenEndPoint (string) - URL of the token endpoint where
+  the Sending Server can exchange a `code` for a bearer token.
```
glpatcern, line 691:
```suggestion
  the Sending Server can exchange a `code` for a bearer token,
  according to the Open ID Connect semantic [RFC6749] [OIDC].
```

spec.yaml
```diff
@@ -466,6 +466,12 @@ components:
                 -----BEGIN PUBLIC KEY-----
                 MII...QDD
                 -----END PUBLIC KEY-----
+        tokenEndPoint:
+          type: string
+          description: >
+            Optional URL path of the Token Exchange endpoint to obtain bearer tokens in exchange for codes.
+            If the `receive-code` capability is exposed, the tokenEndPoint MUST be advertized in the discovery response.
```
```suggestion
            If the `receive-code` capability is exposed, the tokenEndPoint MUST be advertised in the discovery response.
```

spec.yaml
```diff
@@ -854,8 +860,12 @@ components:
...
         grant_type:
           type: string
-          description: Must be set to 'ocm_authorization_code'
-          example: ocm_authorization_code
+          description: Must be set to 'authorization_code'
+          example: authorization_code
+        redirect_uri:
+          type: string
+          description: URI to redirect to after the token is issued
+          example: https://receiver.org/ocm/callback
```
glpatcern, line 875: I guess the callback is not required to be served from the `/ocm` "namespace", as it's a regular OIDC callback?

glpatcern (commented):
OK, I've added a sentence and another question. I guess some details will be cleared once we have a first implementation.

mickenordin:
I think I/we need to look in to https://datatracker.ietf.org/doc/html/rfc6749 more, it feels unclear to me at this point, exactly which flow we should use.

enriquepablo (approved):

spec.yaml
mickenordin: I removed this bit

mickenordin:
Ok, I think my undertanding is cleare now. I have aligned with OAUTH, but just enough so we might get away with outsourcing the token handleing to prexisting oauth libraries without imposing things that are not needed for server to server interactions

IETF-RFC.md
```diff
@@ -1002,6 +1000,83 @@ Additionally, if `protocol.<protocolname>.requirements` includes
...
+    sig1=("@method" "@target-uri" "content-digest" "date"); \
+    created=1730815200; keyid="receiver.example.org#2025"; \
+    alg="rsa-sha256"
+Signature: sig1=:
+    bM2sV2a4oM8pWc4Q8r9Zb8bQ7a2vH1kR9xT0yJ3uE4wO5lV6bZ1cP2rN3qD4tR5hC=:
+
+grant_type=authorization_code&
+client_id=receiver.example.org&
+code=ABCD1234
+```
```
glpatcern, line 1037: How much of this is "example" content vs verbatim content that `MUST` be used in the request?

IETF-RFC.md
```diff
@@ -1002,6 +1000,83 @@ Additionally, if `protocol.<protocolname>.requirements` includes
...
+##  Token Response
+
+If the request is valid and the code is accepted, the Sending Server
+MUST respond with HTTP 200 OK and a JSON object containing the issued
+token:
+
+```
+{
+  "access_token": "8f3d3f26-f1e6-4b47-9e3e-9af6c0d4ad8b",
+  "token_type": "Bearer",
```
glpatcern, line 1056: Isn't that implied? What would be the options?

glpatcern (commented):
Great work Micke, this looks very promising. I have a couple of remaining questions

IETF-RFC.md
KrausMatthias: Regarding the signature, isn't that missing host and content-length? https://github.com/cs3org/OCM-API/blob/develop/IETF-RFC.md#appendix-b-request-signing

IETF-RFC.md
mickenordin: are you referring to the token_type? The idea in the back of my mind is to have enough OAUTH in there, to be able to reuse libraries at least focused on OAUTH, that is why I have tried to retain as much as possible the structure from OAUTH.

IETF-RFC.md
mickenordin: > How much of this is "example" content vs verbatim content that `MUST` be used in the request?

It is meant as example content.

I tried to mimic the style of the OAUTH RFC: https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2

Maybe it is too much? Since it is not _just_ posting a json body, but form encoded request with http-sig, I tried to make it as clear as possible, but maybe I introduced more confusion instead?

IETF-RFC.md
glpatcern: Yep, I referred to the token_type, and you made a very good point, all fine.

IETF-RFC.md
glpatcern: Understood, you are getting expert with the OAuth spec! Let me give it a try to make it clear that it's an example with a more explicit reference to that RFC.

glpatcern (approved):
If you're happy with my additional change, this can be merged.

I have a further question but I'll open a separate issue for that.

PR #295, opened 2025-11-07
Simplified token exchange and related requirements

### LINKED ISSUE #294: Use sharedSecret in the Code flow
As we are now in the process of implementing for real the Code flow and specifying it more accurately as in https://github.com/cs3org/OCM-API/pull/292, I'm thinking that we could slightly change the semantic of a Share payload - at a time where we would not break any implementation.

The point is that `code` and `sharedSecret` are mutually exclusive as it stands now. We could instead drop `code` and signal the fact that we require httpsig and the Code flow by adding the corresponding `use-code` requirement, which would mean that the code to be exchanged is the `sharedSecret` and that barely using the `sharedSecret` as bearer token does not work.

Comments? cc @enriquepablo @mickenordin @KrausMatthias

glpatcern:
This is an attempt to simplify the semantic of the "Code flow", and implements #294.

Notable changes:
* the `receive-code` capability was renamed to `exchange-token`
* the `code` criteria was renamed to `token-exchange`
* the `mfa-enforced` requirement was renamed to `must-use-mfa`
* the `use-code` requirement was renamed to `must-exchange-token`

I believe none of those breaking changes are problematic as to the best of our knowledge there's no implementation yet of the Code flow.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -636,7 +636,7 @@ contain the following information about its OCM API:
   As implementations MUST accept Share Creation Notifications
   to be compliant, it is not necessary to expose that as a
   capability.
-  Example: `["receive-code", "webdav-uri"]`.  The array MAY
+  Example: `["exchange-token", "webdav-uri"]`.  The array MAY
   include for instance:
   _ `"enforce-mfa"` - to indicate that this OCM Server can apply a
   Sending Server's MFA requirements for a Share on their behalf.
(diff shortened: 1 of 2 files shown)


mickenordin (commented):
I think this is a good change

mickenordin:
If we merge this, and publish a lopresti-draft-08 I think we have a good document to do a call for adoption on

glpatcern:
> If we merge this, and publish a lopresti-draft-08 I think we have a good document to do a call for adoption on

I'm afraid we are too late, Thibault sent out the call for adoption!

glpatcern:
I converted to draft because there are more occurrences of `code` throughout the document, I shall fix all of them.

glpatcern:
"Unfortunately", as I looked for all occurrences of `code`, I got into the rabbit hole of the Resource Access chapter. Well, @mickenordin and @enriquepablo please have a look, all comments more than welcome!

mickenordin (approved):
Looks good as far as I can tell

PR #297, opened 2025-11-19
Add functions, roles, and object models to draft

mickenordin:
This updates IETF-RFC.md with a significantly expanded framework for describing OCM using ASCII diagrams in the vain of https://datatracker.ietf.org/doc/rfc3290/ as referenced by https://datatracker.ietf.org/doc/rfc3444/.

It introduces formal definitions of OCM functions, clarifies the roles of Sending and Receiving Servers, and adds detailed object models for Address Books, Contacts, Invites, Providers, Shares, Users, and Resources. The change reorganizes and expands terminology, adds diagrams to illustrate relationships, and alphabetizes and harmonizes defined terms. Legacy duplicated sections are removed or merged into the new structure.


IETF-RFC.md
```diff
@@ -85,60 +85,443 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+## Roles
+
+Open Cloud Mesh defines two distinct roles that an OCM Provider MUST
+take on: the _Sending Server_ role and the _Receiving Server_ role.
+
+### Sending Server
+
+A Sending Server is an OCM Provider that holds Resources and exposes
+APIs to allow access to them.  It allows its users to create _Shares_
+or _Invites_ to give other users access to those Resources. A Sending
```
KrausMatthias: or _Invites_ appears to be duplicated here?

IETF-RFC.md
```diff
@@ -85,60 +85,443 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+an _Address Book_, a _Contact_, an _Invite_, a _Provider_, a _Share_,
+and  a _User_.  However, the following diagrams are provided to clarify
+the concepts and their relationships.
+
+### Address Book
+
+An _OCM Provider_ MAY offer its _Users_ an address book tool, where OCM
+Addresses can be stored over time in a labeled and/or searchable way.
+This decouples the act by which the OCM Address string is passed into
+the Sending Server's database from the selection of the Receiving Party
```
KrausMatthias: Should Receiving Party be _Receiving Party_ ?

IETF-RFC.md
```diff
@@ -85,60 +85,443 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+#### Properties
+
+* __addedDate__: Timestamp of when contact was added
+* __email__: Contact email address (informational)
+* __name__: Human-readable display name
+* __ocmAddress__: Full OCM Address
+* __trusted__: Indicates if contact has been added through the
+               Invite flow (trusted), via direct entry (untrusted)
+               or is blocked
+* __userID__: The identifier of the contact at their OCM Server
```
KrausMatthias: userID is missing in the plot. Why would you need this in addition to the ocmAddress?

IETF-RFC.md
```diff
@@ -85,60 +85,443 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+
+##### Contact States
+
+* __Blocked__: Contact blocked from sending Shares
+* __Trusted__: Established through completed Invite flow
+* __Untrusted__: Added directly without Invite verification
```
KrausMatthias: Not totally convinced of this. What if I want to control who a user may share with? E.g. to avoid accidental data leaks?  Just a bit worried this adds to much detail which is not directly relevant to interoperability.

IETF-RFC.md
```diff
@@ -85,60 +85,443 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+* __status__: Current state (pending, accepted, expired, revoked)
+* __token__: Unique, hard-to-guess string generated by Invite Sender
+             OCM Server
+
+##### Invite States
+
+* __Accepted__: Invite successfully accepted by Invite Receiver
+* __Expired__: Invite no longer valid after a certain time
+* __Pending__: Invite created but not yet accepted
+* __Revoked__: Invite invalidated by Invite Sender before acceptance
```
KrausMatthias: I don't know if I've missed something here as I haven't implemented Invites so far, but why would I store "Non-Pending" invites? Acceptance leads to a trusted entry in the address book. What would I use expired or revoked tokens for?

IETF-RFC.md
```diff
@@ -85,60 +85,443 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+
+#### Properties
+
+* __expiration__: Optional expiration timestamp
+* __name__: Human-readable name of the shared Resource
+* __owner__: OCM Address of the Resource owner
+* __permissions__: Array of granted permissions (read, write, share)
+* __protocol__: Access protocol configuration (webdav, webapp, datatx)
+* __providerId__: Unique identifier for the Share at the provider
+* __requirements__: Array of access requirements (must-use-mfa,
```
KrausMatthias: Why would I need to store those with the (Sent)Share?

IETF-RFC.md
```diff
@@ -85,60 +85,443 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+|    Resource     |
++-----------------+
+```
+
+#### Properties
+
+* __expiration__: Optional expiration timestamp
+* __name__: Human-readable name of the shared Resource
+* __owner__: OCM Address of the Resource owner
+* __permissions__: Array of granted permissions (read, write, share)
```
KrausMatthias: In my eyes permissions currently are protocol specific (at least in the Share Creation Notification) and is stored in the protocol configuration?

IETF-RFC.md
```diff
@@ -166,48 +549,63 @@ characters.
...
+  to the Resource through the Share; similar to "Requesting Party / RqP"
+  in OAuth-UMA, identified by its OCM Address.
+* __Receiving Server__ - The server that:
+  - receives Share Creation Notifications (see below),
+  - actively or passively notifies the receiving user or group of any
+    incoming Share Creation Notification,
+  - acts as an API client, allowing the receiving user to access the
+    Resource through an API (e.g., WebDAV [RFC4918]) of the sending
+    server.
+* __Remote Resource__ - A Resource provided by the Sending Server.
```
KrausMatthias, line 223: A Resource accessible on a Receiving Server via a Share from a Sending Server ?

IETF-RFC.md
```diff
@@ -166,48 +549,63 @@ characters.
...
+  that a Share has been created.
 * __Share Name__ - A human-readable string, provided by the Sending
   Party or the Sending Server, to help the Receiving Party understand
   which Resource the Share grants access to.
 * __Share Permissions__ - protocol-specific allowances granted to the
   Receiving Party on the modes of accessing the Resource.
 * __Share Requirements__ - Protocol-specific restrictions on the modes
   of accessing the Resource.
+* __Shared Resource__ - A Resource shared by an OCM Server, becoming a
+  Remote Resource if accepted by the Invite Receiver OCM Server.
```
KrausMatthias, line 259: Whats the difference to a Resource?

IETF-RFC.md
```diff
@@ -315,6 +703,34 @@ Share Creation Notification from an unknown Sending Party is received
 * the Invite Receiver OCM Server sends the Invite Acceptance Request to
   the Invite Sender OCM Server
 
+```
```
KrausMatthias: Add Users and their actions (Alice and Bob :D )?

KrausMatthias (changes requested):
This is a great improvement! Adds a lot of clarity to the involved concepts.

I think a distinction on what is Binding Specification and what is helpful context should be added.

IETF-RFC.md
mickenordin: I don't see it?

mickenordin: do you mean it should be invites, below, without italics?

KrausMatthias: Maybe I'm overly picky. In the sentence 

> It allows its users to create _Shares_ **or _Invites_** to give other users access to those Resources.

Invites don't have anything to do with Resource Access. Also creating invites is not optional there, but is in the next sentence.

IETF-RFC.md
mickenordin: I was think about this question when writing it, and struggled abit, since the same model is used for both local and remote users. But maybe it makes sense to drop the local uid and just use the ocm address.

IETF-RFC.md
mickenordin: I could remove the states and just mention in text that an implementation MAY treat invites differently from direct entry.

IETF-RFC.md
mickenordin: Same here, maybe just remove the states then?

IETF-RFC.md
mickenordin: These are part of a share object according to the spec, and at least as sending server you need to for example keep track of what permissions was mandated for the share, no?

IETF-RFC.md
mickenordin: These two (this and your comment above) have been part of the definitions for a long time, I just alphabetized them. But I take your point that there is a lot of redundancy in definitions, and I was also pondering that when sorting them.

IETF-RFC.md
MahdiBaghbani: I'm also in favor of using the OCM Adress in here

mickenordin: All our api's use userID and/or (userID and provider):

* https://github.com/cs3org/OCM-API/blob/7821ae8dc982fe9e752cd0c84f75dc1015460649/spec.yaml#L815

* https://github.com/cs3org/OCM-API/blob/7821ae8dc982fe9e752cd0c84f75dc1015460649/spec.yaml#L785

So maybe we ought to do that here as well?

mickenordin: I implemented invites today, in my toy app, and there is alot of this: https://code.smolnet.org/micke/amity/src/commit/2c26625ee0d45ab42bcaf249fa9314eab08d2228/app/paths/invite_accepted.py#L32

https://code.smolnet.org/micke/amity/src/commit/2c26625ee0d45ab42bcaf249fa9314eab08d2228/app/paths/accept_invite.py#L96

So I think I am in favor of using userID instead of the ocm-address

KrausMatthias:
I think my main concern is still what is strictly necessary for interoperability, and what should be left open to implementations (while guidance regarding the concepts is still extremely helpful there). 

So I'm not after removing details but clearly separating, "this must what a share payload looks like with these semantics so it works with other implementations" and "this is what you might keep track of in your implementation to build a working service, but doing it differently would still work". 

So maybe we could have two separate sections? "Protocol Specification" and "Context and Implementation Guidance" ?

glpatcern:
> I think my main concern is still what is strictly necessary for interoperability, and what should be left open to implementations (while guidance regarding the concepts is still extremely helpful there).

Overall I'm quite sympathetic with this comment. I appreciate the overall effort to "canonicalize" the implementations, but it has to be clear what is prescribed and what is guidance. I did not have time yet to review this PR in details and I thank both of you for the combined effort!

IETF-RFC.md
glpatcern: Indeed, we also store in our data model an `opaque_user_id` and an `idp` as opposed to the OCM Address, I think it makes sense from a modelling perspective.

IETF-RFC.md
```diff
@@ -315,6 +683,34 @@ Share Creation Notification from an unknown Sending Party is received
...
+|   Added     |                                  |   Invite    |
++-------------+                                  +-------------+
+                                                        |
+                                                        v
+                                                 +-------------+
+                                                 |  Contact    |
+                                                 |   Added     |
+                                                 +-------------+
+```
+
```
glpatcern, line 404: Here I'd either include the mermaid-formatted Sequence Diagrams we have in the repo or refer to them, as opposed to try and draw them.

IETF-RFC.md
```diff
@@ -166,48 +529,63 @@ characters.
...
+  being granted, including but not limited to: a file or folder, a video
+  call, a contact, a printer queue, etc.
+* __Sending Gesture__ - A user interface interaction from the Sending
+  Party to the Sending Server, conveying the intention to create a
+  Share.
+* __Sending Party__ - A person or party who is authorized to create
+  Shares; similar to "Resource Owner" in OAuth [RFC6749], identified by
+  its OCM Address.
+* __Sending Server__ - The server that:
+  - holds the Resource ("file server" or "Entreprise File Sync and Share
```
glpatcern: I'd remove the `file server or EFSS` example, or explicitly state `for instance, a file server`.

IETF-RFC.md
```diff
@@ -85,60 +85,423 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+
+### Receiving Server
+
+A Receiving Server is an OCM Provider that receives _Share_ Creation
+Notifications from Sending Servers, notifies its users about incoming
+_Shares_, and acts as an API client to allow its users to access Remote
+Resources. It MAY provide it's users with an _Address Book_ of
+_Contacts_ and the ability to accept _Invites_.
+
+## Object models
```
glpatcern: I'm not so convinced that this section should be at the very beginning of the Draft. What do other people think?

glpatcern (commented):
I started to give a full review and already have a few comments, possibly more to come, but overall I think it's an effort in the good direction

IETF-RFC.md
mickenordin: We can put it in an appendix, maybe that is better?

glpatcern: Yes, that seems better indeed

mickenordin:
> > I think my main concern is still what is strictly necessary for interoperability, and what should be left open to implementations (while guidance regarding the concepts is still extremely helpful there).
> 
> Overall I'm quite sympathetic with this comment. I appreciate the overall effort to "canonicalize" the implementations, but it has to be clear what is prescribed and what is guidance.

I think many of these concerns are addressed now, by moving it to an appendix and stating that it is non-normative.

IETF-RFC.md
```diff
@@ -85,60 +85,96 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
-  another reason.
-* __Sharing User__ - A user providing access to a Resource through a
-  Share.
-* __FQDN__ - Fully Qualified Domain Name, such as `"cloud.example.com"`.
-* __OCM Server__ - A server that supports OCM.
-* __OCM API Discovery__ - Process of evaluating properties of a Remote
-  Resource, after establishing contact with an OCM Server.
-* __Discovering Server__ - A server that tries to obtain information in
-  OCM API Discovery.
+Open Cloud Mesh defines distinct functions, it is not neccessary for an
```
```suggestion
Open Cloud Mesh defines distinct functions.  It is not necessary for an
```

IETF-RFC.md
```diff
@@ -85,60 +85,96 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+## Roles
+
+Open Cloud Mesh defines two distinct roles that an OCM Provider MUST
+take on: the _Sending Server_ role and the _Receiving Server_ role.
+
+### Sending Server
+
+A Sending Server is an OCM Provider that holds Resources and exposes
+APIs to allow access to them.  It allows its users to create _Shares_
+to give other users access to those Resources. A Sending Server MAY
```
```suggestion
to give other users access to those Resources.  A Sending Server MAY
```

IETF-RFC.md
```diff
@@ -85,60 +85,96 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+provide its users with the ability to generate _Invites_ to establish
+contact with other users on other OCM Providers. When doing so it MAY
```
```suggestion
contact with other users on other OCM Providers.  When doing so it MAY
```

IETF-RFC.md
```diff
@@ -85,60 +85,96 @@ NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
...
+be limited to a set of trusted OCM Providers, for instance those in the
+same _Federation_.
+
+
+### Receiving Server
+
+A Receiving Server is an OCM Provider that receives _Share_ Creation
+Notifications from Sending Servers, notifies its users about incoming
+_Shares_, and acts as an API client to allow its users to access Remote
+Resources. It MAY provide it's users with an _Address Book_ of
```
```suggestion
Resources.  It MAY provide its users with an _Address Book_ of
```

IETF-RFC.md
```diff
@@ -218,8 +268,9 @@ characters.
     or supported capabilities.
 * __WAYF Page__ - A Where-Are-You-From page is a discovery service used
   to identify the OCM Server of an Invite Receiver.
-* __Directory Service__ - A third-party service that exposes a list of
-  trusted OCM Servers.
+
+In Appendix D, an object model is presented as a non-normative guide for
+implementators to understand the relationships between these terms.
```
```suggestion
implementers to understand the relationships between these terms.
```

IETF-RFC.md
```diff
@@ -1394,6 +1438,337 @@ format:
 }
 ```
 
+
+
+# Appendix D: Object models
+
+An implementor of OCM MAY choose any internal object model to represent
+an _Address Book_, a _Contact_, an _Invite_, a _Provider_, a _Share_,
+and  a _User_.  However, the following diagrams are provided to clarify
```
```suggestion
and a _User_.  The following diagrams are provided to clarify
```

IETF-RFC.md
```diff
@@ -1394,6 +1438,337 @@ format:
...
+## Address Book
+
+An _OCM Provider_ MAY offer its _Users_ an address book tool, where OCM
+Addresses can be stored over time in a labeled and/or searchable way.
+This decouples the act by which the OCM Address string is passed into
+the Sending Server's database from the selection of the _Receiving
+Party_ in preparation for Share Creation.
+
+The Address Book entity maintains a collection of contacts for a user
+within the OCM provider. It serves as the primary mechanism for managing
```
glpatcern: Here and in the following we should still follow the IETF convention of two-spaces-after-full-stop

glpatcern (changes requested):
I think this looks good now, the requested changes are just spelling mistakes.

glpatcern (dismissed):
I'm happy with this now. @KrausMatthias what about you?

KrausMatthias (commented):
Should we distinguish between Sent Shares and Received Shares in the Object models?

Otherwise I think it's perfect :)

KrausMatthias (approved):

MahdiBaghbani (dismissed):
I've read it now, It's great, nice ascii flowcharts! Thanks Micke

glpatcern:
> Should we distinguish between Sent Shares and Received Shares in the Object models?

True that in a real implementation (e.g. in Reva) those two types are expected to be quite different, at least because in the Received Share the remote URI has to be stored. We may want to emphasize that.

mickenordin:
> Should we distinguish between Sent Shares and Received Shares in the Object models?
> 
> Otherwise I think it's perfect :)

This is a great point, I will add a 'direction' to the model.

mickenordin:
I added a direction to the share model now

IETF-RFC.md
```diff
@@ -1419,32 +1416,361 @@ adhere to the following format:
...
++-----------------+
+```
+
+### Properties
+
+* __direction__: Whether the Share is incoming or outgoing
+* __expiration__: Optional expiration timestamp
+* __name__: Human-readable name of the shared Resource
+* __owner__: OCM Address of the Resource owner
+* __protocol__: Access protocol configuration (webdav, webapp, datatx)
```
```suggestion
* __protocol__: Access protocol name and details (webdav, ssh, webapp, ...)
```

IETF-RFC.md
```diff
@@ -1419,32 +1416,361 @@ adhere to the following format:
...
+| - protocols{}    |  | - invite-wayf    |   | - http-signatures|
++------------------+  | - invites        |   | - invite         |
+       |              | - webdav-uri     |   | - token-exchange |
+       |              +------------------+   +------------------+
+       | supports
+       v
++------------------+
+|   Protocols      |
++------------------+
+| - datatx         |
```
```suggestion
| - ssh            |
```

IETF-RFC.md
```diff
@@ -1419,32 +1416,361 @@ adhere to the following format:
...
+store the discovered information about federation peers or other remote
+OCM Providers.
+
+```
+            +-----------------------+
+            |      Provider         |
+            |    (OCM Server)       |
+            +-----------------------+
+            | - apiVersion          |
+            | - enabled: bool       |
```
```suggestion
            | - enabled             |
```

glpatcern (changes requested):
The changes below are trivial fixes + the removal of `datatx`.
Otherwise, concerning the `direction`, I'm not so sure it helps at this level of abstraction. As a matter of fact, the `protocols` encapsulate the access details, including whether the share is incoming or sent, and a real model would have other more significant differences (cf. for example in Reva the model for a [sent share](https://github.com/cs3org/reva/blob/master/pkg/share/manager/sql/model/model.go#L161-L175) vs an [incoming share](https://github.com/cs3org/reva/blob/master/pkg/share/manager/sql/model/model.go#L186-L201)).
As such, I'd vote for removing the `direction` (stating that it's part of the access protocols), or for further detailing the model as secondary option.

mickenordin:
> The changes below are trivial fixes + the removal of `datatx`. Otherwise, concerning the `direction`, I'm not so sure it helps at this level of abstraction. As a matter of fact, the `protocols` encapsulate the access details, including whether the share is incoming or sent, and a real model would have other more significant differences (cf. for example in Reva the model for a [sent share](https://github.com/cs3org/reva/blob/master/pkg/share/manager/sql/model/model.go#L161-L175) vs an [incoming share](https://github.com/cs3org/reva/blob/master/pkg/share/manager/sql/model/model.go#L186-L201)). As such, I'd vote for removing the `direction` (stating that it's part of the access protocols), or for further detailing the model as secondary option.

Fixed the nits, and removed direction

glpatcern (dismissed):

mickenordin:
I had to fix a merge confict, and that made the reviews stale... Can you please re-review?

glpatcern (dismissed):

glpatcern (commented):
Yep, I got there was a conflict, but are you sure this is now the final state? It seems the paragraph about the `Functions` is gone

mickenordin:
> Yep, I got there was a conflict, but are you sure this is now the final state? It seems the paragraph about the `Functions` is gone

I added back functions and roles, does it look ok now?

glpatcern (approved):
Yes!

PR #298, opened 2025-11-25
add: redoc script, load it in document viewer

MahdiBaghbani:


mickenordin (approved):
good idea

glpatcern:
Nice, but can we source the redoc javascript from somewhere else, as opposed to keep a copy here?

MahdiBaghbani:
@glpatcern fixed that as well, need to see if it loads, created new PR #229

PR #304, opened 2025-11-26
Adds a schemas directory and one schema so far, for the OCM Discovery response in section 5

lisad:
I'm not sure whether these are going to be normative or not, but I worked up a schema for section 5 to start with

### DIFF
--- a/schemas/ocm-discovery.json
+++ b/schemas/ocm-discovery.json
@@ -1,0 +1,1 @@
+ocm-discovery.json
--- a/schemas/ocm-discovery.json
+++ b/schemas/ocm-discovery.json
@@ -1,1 +1,101 @@
-ocm-discovery.json
+{
+  "title": "OCM API Discovery",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "enabled": {
+      "type": "boolean"
+    },
+    "apiVersion": {
+      "type": "string"
+    },
+    "endPoint": {
+      "type": "string", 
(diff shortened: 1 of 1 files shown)


glpatcern:
Hi Lisa, great to see you starting on this, but it seems you may have not committed the schema itself or did I miss something?

lisad:
I put a file called "ocm-discovery.json" inside a new folder called "schemas"?   there will be several separate schemas if I continue with the rest of the spec.  I made the schema a separate file for a number of reasons.  
1) it's easier for us not to screw up the syntax when editing it (trying to editing valid JSON inside MD is error-prone for me), we can even add a github test action to confirm schemas are valid
2) folks will eventually be able to refer directly to these schemas when checking their implementations - some folks do this in CI/CD tests
3) I wasn't sure we'd want to have them in the spec, but if we do, we can still import the schema into the spec  - see for example how Rohan Mahy uses {::include} within https://github.com/rohanmahy/mimi-message-status/blob/main/draft-mahy-mimi-message-status.md to merge the schemas into the draft.

MahdiBaghbani:
@lisad I can see the file, but its contents are empty. Like Giuseppe, I assumed it hadn't been committed yet

lisad:
Oh sorry I get it now!  Fixed.

glpatcern (approved):
This seems a very good direction for all the schemas! As a matter of fact, currently the schemas are declared in the OpenAPI `spec.yaml` file, but that's not linked in the Draft, and instead we can both link them and enforce them in CIs.

PR #305, opened 2025-12-03
Add RFC 9421 support via rfc-http-sig capability and publicKeys array

mickenordin:
* Add rfc-http-sig capability to advertise RFC 9421 support
* Add publicKeys array with algorithm field for HTTP Message Signatures
* Deprecate publicKey field in favor of publicKeys
* Reference IANA HTTP Signature Algorithms Registry
* Maintain backward compatibility with draft-cavage signatures used by Nextcloud

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -657,6 +657,9 @@ contain the following information about its OCM API:
   for a short-lived bearer token.
   _ `"invite-wayf"` - to indicate that this OCM Server exposes a WAYF
   Page to facilitate the Invite flow.
+  _ `"http-sig"` - to indicate that this OCM Server supports
+  [RFC9421] HTTP Message Signatures and advertises public keys in the
+  `publicKeys` array for signature verification.
 * OPTIONAL: criteria (array of string) - The criteria for accepting a
   Share Creation Notification.
   As all Receiving Servers SHOULD require the use of TLS in API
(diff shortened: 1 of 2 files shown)


IETF-RFC.md
```diff
@@ -657,6 +657,9 @@ contain the following information about its OCM API:
   for a short-lived bearer token.
   _ `"invite-wayf"` - to indicate that this OCM Server exposes a WAYF
   Page to facilitate the Invite flow.
+  _ `"rfc-http-sig"` - to indicate that this OCM Server supports
```
```suggestion
  _ `"http-sig"` - to indicate that this OCM Server supports
```

IETF-RFC.md
```diff
@@ -674,19 +677,37 @@ contain the following information about its OCM API:
...
     Example: https://my-cloud-storage.org/ocm#signature
-  - REQUIRED publicKeyPem (string) - PEM-encoded version of the public
-    key.
+  - REQUIRED publicKeyPem (string) - PEM-encoded RSA public key for
+    draft-cavage signatures.
     Example:
     "----BEGIN PUBLIC KEY----\n...\n----END PUBLIC KEY----\n"
+* OPTIONAL: publicKeys (array of objects) - Array of public keys for
+  [RFC9421] HTTP Message Signatures.
+  Servers advertising the "rfc-http-sig" capability MUST provide this
```
```suggestion
  Servers advertising the "http-sig" capability MUST provide this
```

IETF-RFC.md
```diff
@@ -674,19 +677,37 @@ contain the following information about its OCM API:
...
+  field. Clients SHOULD prefer [RFC9421] signatures when this capability
+  is present. Each object in the array MUST contain:
+  - REQUIRED keyId (string) - Unique identifier for this key in URI
+    format. Hostname MUST match the discovery endpoint hostname.
+    Example: https://my-cloud-storage.org/ocm#key-1
+  - REQUIRED publicKeyPem (string) - PEM-encoded public key for [RFC9421]
+    signatures.
```
```suggestion
  - REQUIRED publicKeyPem (string) - PEM-encoded public key for
    [RFC9421] signatures.
```

IETF-RFC.md
```diff
@@ -674,19 +677,37 @@ contain the following information about its OCM API:
...
+    Example:
+    "----BEGIN PUBLIC KEY----\nMCowBQYDK2VwAyEA...\n----END PUBLIC KEY----\n"
```
```suggestion
    "----BEGIN PUBLIC KEY----\nMCowBQYDK...\n----END PUBLIC KEY----\n"
```

spec.yaml
```diff
@@ -420,9 +420,11 @@ components:
               - protocol-object
               - invites
               - exchange-token
+              - rfc-http-sig
```
```suggestion
              - http-sig
```

spec.yaml
```diff
@@ -420,9 +420,11 @@ components:
...
           example:
             - webdav-uri
             - protocol-object
+            - rfc-http-sig
```
```suggestion
            - http-sig
```

spec.yaml
```diff
@@ -442,26 +444,61 @@ components:
...
           type: object
+          deprecated: true
           description: >
-            The signatory used to sign outgoing request to confirm its origin.
-            The 
-            signatory is optional but it MUST contain `id` and `publicKeyPem`.
+            DEPRECATED: Use publicKeys array instead for RFC 9421 support.
+            Legacy field for draft-cavage HTTP Signatures (RSA only).
+            Maintained for backward compatibility with existing deployments.
+            The signatory is optional but it MUST contain `keyId` and `publicKeyPem`.
```
glpatcern, line 452: As commented in person, I'm tempted to say we entirely drop this as opposed to deprecate it, despite we know there is one implementation. I'd like to see what other people think about this.

glpatcern (changes requested):
Some minor changes to fit in the width and to just refer to `http-sig` as capability.

spec.yaml
MahdiBaghbani: I also agree that we should drop it, true that Nextcloud is still using this but in case of an update, they can still backport it.

mickenordin: Since we change the key, and allow arbitrary fields it will not impact Nextcloud if we remove it, but we have a lot of other deprecated things in there already and it is good practice to deprecate before removing I think. But not a strong opinion, I am fine with just removing it if that is what people want

glpatcern (approved):
So it seems there's no strong opinion either way. Maybe let's bring the topic of deprecation to the mailing list? Otherwise this looks good for me

PR #307, opened 2025-12-08
Feat: introduce accessType property in a share and remove the datatx "protocol"

glpatcern:
Following discussions last week at the SIG-CISS meeting, this PR implements the following:
* Introduction of an `accessType` property inside the `webdav` protocol, with values `remote` and `datatx` (defaults to `remote`)
* Removal of `datatx` as protocol, in favor of the above

The rationale is that data transfers are "orthogonal" to the underlying transport protocol, especially if we think about other possible protocols (another companion PR is coming next).

Some comments:
* I have _removed_ `datatx` as protocol on the basis that it's not used in any production setup (as far as I know, the only implementation is in [Reva](https://github.com/cs3org/reva/blob/master/internal/grpc/services/gateway/ocmshareprovider.go) but I may be wrong). If people prefer an official deprecation, that's also fine.
* The new property remains optional for backwards compatibility, and it is meant to advice the recipient about how to consume the share. Should it be more prescriptive instead?

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -623,11 +623,6 @@ contain the following information about its OCM API:
       endpoint.  This value is provided for documentation
       purposes, and it SHOULD NOT be intended as a prefix
       for share requests.
-    - datatx (string) - The top-level path used for data transfers.
-      This value is provided for documentation purposes,
-      and it SHOULD NOT be intended as a prefix.  In
-      addition, implementations are expected to execute
-      the transfer using WebDAV [RFC4918] as the wire protocol.
     - Any additional protocol supported for this Resource type MAY be
       advertised here, where the value MAY correspond to
       a top-level URI to be used for that protocol.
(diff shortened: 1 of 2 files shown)


mickenordin (dismissed):
Very good, one thing I wonder is if we should mention in the text is that datatx without an expiration is intended for `sync` like use cases?

glpatcern:
> Very good, one thing I wonder is if we should mention in the text is that datatx without an expiration is intended for `sync` like use cases?

Good point, why not. And even reinforce that `datatx` with expiration is targeting moving of dataset and deleting the source, e.g. if the source is a cache.

What about deprecating the `datatx` protocol? I'm slightly in favor of removing it without deprecation but I'm really open to opinions. In particular, I invite @lisad and @thibmeu to comment here about best practices for this change.

mickenordin (approved):

PR #308, opened 2025-12-08
Feat: add support for ssh

### LINKED ISSUE #277: RFE: Support for ssh and scp/sftp
Specific inclusion of ssh, including scp and/or sftp, into the ocm specification could be beneficial for many uses cases. This could include ocm account linking, where a receiving party then is identified by one or multiple public keys.

glpatcern:
This PR is stacked on top of #307 and defines support for an `ssh` protocol, either for remote access (e.g. via `sshfs`) or as a data transfer protocol via `scp`. Fixes #277.

In the process of writing it, I realized that the Draft did not include any explanation of the Response payload of a share (!), so that's included as well with a separate commit, which I'd like not to squash when the PR gets merged.

An important issue when dealing with ssh/scp is that a local account for a remote user at the sharing server end has to be used, at variance with web-based protocols where remote users do not require it. @sunetfreitag how is FileSender going to approach this?

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -866,11 +866,12 @@ To create a Share, the Sending Server SHOULD make a HTTP POST request
     third-party service to execute the data transfer on their behalf.
   - REQUIRED uri (string)
     The full address to be used for ssh or scp access, in the form
-    `username@host.fqdn:port/resource/path`, where authentication is
-    expected to take place via public/private key: the Receiving Server
-    MUST reply to such a Share Creation Notification by sending back
-    their public key, for the Sender Server to authorize access to the
-    Resource.
+    `username@host.fqdn:port/resource/path`, where the `username` is
+    chosen by the Sending Server and does not necessarily need to match
+    the recipient's OCM Address.  Authentication is expected to take
+    place via public/private key: the Receiving Server MUST reply to
+    such a Share Creation Notification by sending back their public
+    key, for the Sender Server to authorize access to the Resource.
 
 ## Response
 
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -607,22 +607,23 @@ contain the following information about its OCM API:
     support at least `webdav`,
     any other combination of Resources and protocols is
     optional.  Example:
-    `json
+    ```json
             {
               "webdav": "/remote/dav/ocm/",
               "webapp": "/app/ocm/",
               "talk": "/apps/spreed/api/"
             }
-            `
+    ```
     Fields:
     - webdav (string) - The top-level WebDAV [RFC4918] path at this
       endpoint.  In order to access a Remote Resource, implementations
-      MAY use this path as a prefix, or as the full path (see sharing
-      examples).
+      SHOULD use this path as a prefix (see sharing examples).
     - webapp (string) - The top-level path for web apps at this
-      endpoint.  This value is provided for documentation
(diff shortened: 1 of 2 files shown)


mickenordin (dismissed):
Yes, very good.

MahdiBaghbani (dismissed):
This is a good addition!
I wonder if the wrong response payload (code 200?) should result in red status in the OCM Stub tests

glpatcern:
> This is a good addition! I wonder if the wrong response payload (code 200?) should result in red status in the OCM Stub tests

Well, it should indeed! Maybe "orange" if you have such a concept

mickenordin (approved):

IETF-RFC.md
```diff
@@ -768,47 +764,54 @@ To create a Share, the Sending Server SHOULD make a HTTP POST request
...
+    browsing.  - `datatx` signals the recipient that the resource is
+    available for data transfer.  If no expiration is given, the share
+    is suitable e.g. for sync use-cases, whereas if an expiration date
+    is set, the above clause MAY apply and the recipient SHOULD notify
+    the sender upon completing the data transfer, in order to ease
+    cache operations on the Sending Server.  The recipient MAY delegate
+    a third-party service to execute the data transfer on their behalf.
   - REQUIRED uri (string)
     A URI to access the Remote Resource.  The URI
     SHOULD be relative, in which case the prefix
```
mickenordin, line 817: Wait a minute! I think, that we need more info here. We need user@host:port/resource/path in the fully resolved uri, so either we must include a user part in the discovery alongside host and port (both mandatory imho, no fallback to port 22) OR we make uri always absolute here, including user part and resource path. Maybe the sane thing is to mandate a user part in the discovery, like `ocm@amity.smolnet.org:9022` but then we are limited to a single user for all ssh access (simliar to how many git implementations work with git@github.com for example)

mickenordin (changes requested):

IETF-RFC.md
mickenordin: so what do we like: `ocm@amity.smolnet.org:9022/micke/my_cool_file.md` or `micke@amity.smolnet.org:9022/my_cool_file.md`?

mickenordin: Or am I over thinking this? Should the username come from the local part of the address in the share? I was thinking, we may want to separate the ocm user name from the ssh access user.

MahdiBaghbani: Git way makes more sense to me, in a sense that managing the actual ssh user is easier for EFSS

glpatcern: I think the details on how to access via ssh - including the local account to be used for remote users, which typically is not a concern with web-based protocols! - are yet to be fleshed out, and we better stay as generic as possible. I'd love to hear from the actual use case @sunetfreitag had in mind.

Otherwise, I agree we should include the port without fall-back default. I'll fix that.

glpatcern:
I've rebased this and made it clear that for ssh access we need a full address in the form `username@host.fqdn:port/resource/path`. Whether the username is receiver-specific or totally generic (similar to `git@github.com`) is left to the implementation.

mickenordin (approved):
Ok, good, I started to implement this in amity and figured out I had to have a complete virtual file system if all I had was the public key to distinguish between shares, since that may well be reused between multiple shares across multiple users. Now, that turned out pretty good, so I may keep it, but at least now we don't need to do it that way :)

PR #309, opened 2025-12-09
Prescribe use of JWS for Directory Service

glpatcern:
Supersedes #290, see there for the discussion.

I deliberately left the distribution of the key to validate the signature out of the spec, but we might want to suggest something.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -1359,7 +1359,7 @@ request.
 A third-party Directory Service is a back-end service used to federate
 multiple OCM Servers and facilitate the Invite flow.  It is expected to
 expose, via anonymous HTTPS GET, a signed JWS document [RFC7515], where
-the signature MUST be made available offline and the payload MUST
+the signing key MUST be made available offline and the payload MUST
 adhere to the following format:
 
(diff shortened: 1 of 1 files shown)


mickenordin (dismissed):

IETF-RFC.md
```diff
@@ -1355,8 +1358,9 @@ request.
 
 A third-party Directory Service is a back-end service used to federate
 multiple OCM Servers and facilitate the Invite flow.  It is expected to
-expose, via anonymous HTTP GET, a JSON document with the following
-format:
+expose, via anonymous HTTPS GET, a signed JWS document [RFC7515], where
+the signature MUST be made available offline and the payload MUST
```
KrausMatthias: the signing key instead of signature?

KrausMatthias: One could also think about optionally only accepting signed requests of servers within the directory?

glpatcern: Absolutely, that would be at the discretion of the site administrator. Should I add an explicit paragraph, do you want to suggest something?

KrausMatthias: I guess you are probably right with keeping it simple for now

mickenordin (approved):

PR #310, opened 2025-12-10
Fix validation error for tokenEndPoint

mickenordin:
The tokenEndPoint does not pass validation:

spec.yaml: Validation Error: Path parameter 'tokenEndPoint' for 'post' operation in '/{tokenEndPoint}' was not resolved (using openapi-spec-validator)

This is an attempt to fix that by removing the endpoint and adding the information under the discovery instead, since this endpoint is discovered rather than fixed, I _think_ this is how it should be done.

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -433,30 +433,18 @@ components:
             Optional URL path of the Token Exchange endpoint to obtain bearer tokens in exchange for codes.
             If the `exchange-token` capability is exposed, the tokenEndPoint MUST be advertised in the discovery response.
 
-
             **Token Exchange API:**
-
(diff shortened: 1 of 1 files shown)


glpatcern:
Interestingly, the CI does not fail. How did you see the failed validation?

And maybe you could remove the extra empty lines?

mickenordin:
> Interestingly, the CI does not fail. How did you see the failed validation?
 
I use the tool `openapi-spec-validator` (`pipx install openapi-spec-validator`)

> And maybe you could remove the extra empty lines?

Certainly!

glpatcern (approved):
Looks good, we could integrate the `openapi-spec-validator` tool in the CI then

PR #311, opened 2025-12-12
Make recipientPublicKey plural

mickenordin:
I order for server-to-server interactions to work over ssh, the recipient OCM server needs to have a private key, that can be used for access, and an implementation may ask the user to upload a key pair for this purpose, or generate one on behalf of the user.

However a user may also want to access the share directly using other means, or an enterprise may want to enable third-party-copy and for this reason it will be useful to be able to reply with multiple public keys, used for different purposes.


glpatcern (approved):
I see the scope of this, ideally we want some informed opinions on how this is being modeled.

PR #312, opened 2025-12-13
Add `ssh` to supported protocols

mickenordin:
Minor spec bug, where ssh was missing from supported protocols


glpatcern (approved):
Right, missed it in my PR!

PR #316, opened 2025-12-27
Change my-cloud-storage.org -> cloud.example.org

mickenordin:
According to the Internet-Draft Author Resources at:

https://authors.ietf.org/en/example-addresses

and RFC 2606: https://datatracker.ietf.org/doc/rfc2606/

we should use appointed example domain names and for these examples a second level example domain is appropriate, and .org is used here as the tld as in the previous examples.


glpatcern (dismissed):
Yeah, I had this in the back of my mind for some time :-)

glpatcern (approved):

PR #318, opened 2026-01-11
Add JSContact extension to IANA Considerations

mickenordin:
Register ocmAddress type and ietf.org:ocmAddresses property for storing OCM addresses in JSContact Cards (RFC 9553).


glpatcern (approved):
This is a nice addition!

PR #319, opened 2026-01-12
JWKS: Remove publicKeys in favor of RFC7515.

### LINKED ISSUE #313: Adopt JWS format for exposing public keys
A suggestion brought forward by @thibmeu at yesterday's call is to adopt the JOSE framework and in particular (as I understand it) the JWS format from [RFC 7515](https://datatracker.ietf.org/doc/rfc7515) to expose the public key(s) to be used to validate httpsig-signed messages.

Indeed, I see the typical practice is to expose a `https://ocm-server.org/.well-known/jwks.json` file that would have a format such as:
```
{
  "keys": [
    {
      "kty": "<key type, e.g. RSA>",
      "use": "sig",
      "kid": "key-identifier-2025-12",
      "alg": "RS256",
      "n": "...",
      "e": "...",
    }
  ]
}
```
(or similar for an EC key)

For OCM, this would mean:
* Entirely dropping the pubkey/pubkeys fields in the OCM discovery endpoint.
* Stating that the capability `http-sig` and corresponding `request-http-signature` criteria entail exposing/discovering the signing key from the above endpoint.
We could also consider to introduce a `must-use-http-sig` requirement, rather than having the criteria.

@mickenordin you've most recently implemented signatures in your "toy" system, what do you think?

mickenordin:
Fixes: #313

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -729,7 +729,7 @@ contain the following information about its OCM API:
   for a short-lived bearer token.
   _ `"http-sig"` - to indicate that this OCM Server supports
   [RFC9421] HTTP Message Signatures and advertises public keys in the
-  format specified by [RFC7515] at the `/.well-known/jwks.json`
+  format specified by [RFC7517] at the `/.well-known/jwks.json`
   endpoint for signature verification.
   _ `"invites"` - to indicate the server would support acting as an
   Invite Sender or Invite Receiver OCM Server.  This might be useful
@@ -1363,7 +1363,14 @@ June 2007.
 https://datatracker.ietf.org/html/rfc6749)", October 2012.
 
 [RFC7515] Jones, M., Bradley, J., Sakimura, N., "[JSON Web Signature
-(JWS)](https://datatracker.ietf.org/doc/html/rfc7515), May 2015."
+(JWS)](https://datatracker.ietf.org/doc/html/rfc7515)", May 2015.
+
+[RFC7517] Jones, M., "[JSON Web Key (JWK)](
+https://datatracker.ietf.org/doc/html/rfc7517)", May 2015.
+
+[RFC8032] Josefsson, S., Liusvaara, I., "[Edwards-Curve Digital
+Signature Algorithm (EdDSA)](
+https://datatracker.ietf.org/doc/html/rfc8032)", January 2017.
 
 [RFC8174] Leiba, B. "[Ambiguity of Uppercase vs Lowercase in RFC 2119
 Key Words](https://datatracker.ietf.org/html/rfc8174)", May 2017.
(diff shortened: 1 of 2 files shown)


glpatcern (commented):
This is already good, do you think we could provide an example of how to use the jwks.json? Also, at this point I'm ok to keep the deprecated part, but I'd remove the now-fully-outdated example.

mickenordin:
> This is already good, do you think we could provide an example of how to use the jwks.json? Also, at this point I'm ok to keep the deprecated part, but I'd remove the now-fully-outdated example.

I removed the deprecated example now. Where would be a good place to put an example, and how would you like it to be? An example json with the various fields?

glpatcern:
> I removed the deprecated example now. Where would be a good place to put an example, and how would you like it to be? An example json with the various fields?

Good question: what about an Appendix as it was done with the previous example? Then yes I'd include an example JSON like the one in #313, and a code snippet from your own toy implementation (it's python IIRC right?) to 1) encode and add a signature at the sender and 2) validate it at the receiver.

mickenordin:
> > I removed the deprecated example now. Where would be a good place to put an example, and how would you like it to be? An example json with the various fields?
> 
> Good question: what about an Appendix as it was done with the previous example? Then yes I'd include an example JSON like the one in #313, and a code snippet from your own toy implementation (it's python IIRC right?) to 1) encode and add a signature at the sender and 2) validate it at the receiver.

Good thing you mentioned appendices.. We hade appendix b with draft-cavage signature examples. I rewrote that, and I hope it is something what you wanted. A lot of python code didn't look good so a step by step algorithm turned out better I think.

glpatcern:
> Good thing you mentioned appendices.. We hade appendix b with draft-cavage signature examples. I rewrote that, and I hope it is something what you wanted. A lot of python code didn't look good so a step by step algorithm turned out better I think.

Ah yes, that appendix indeed was also obsoleted now, great that you rewrote it.

I think the text and the step by step part is good but the ~~hacker~~programmer that is in me would still like to see some minimal code, let me give you an example.

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
-    `content-digest` and `host`,
-  - `signature` is the encrypted version of the
-    signature.
+```
+"@method": POST
+"@target-uri": https://receiver.example.org/ocm/shares
+"content-digest": sha-256=:LkpHyFOVbBDPxc7YbHDOWNzAv88qWuVfLNf4TUf9Uo8=:
+"@signature-params": ("@method" "@target-uri" "content-digest")\
+  ;created=1618884475;keyid="sender.example.org#key1";alg="ed25519"
+```
```
glpatcern, line 1480: So for the examples, is there maybe some library that takes a `method`, `target-uri` etc and builds this?
Ideally `base = build_sig_base(...)`?

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
 
-Here is an example of how to verify the signature using the headers,
-the signature and the public key:
+Sign this base using, for example Ed25519 ([RFC8032]) to produce the
+signature, then add headers:
```
glpatcern: And here I'd still put a (python) snippet to compute the signature of a `base` where the `base` is as obtained just above, then yes just say "append the `Signature-Input` and `Signature` headers, for example:"

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
 
-## Validating the payload
+1. Extract the provider domain from the `sender` field in the
+   request body
+2. Fetch the public key from
+   `https://<provider-domain>/.well-known/jwks.json`
+3. Extract `keyid` from `Signature-Input` header and find the key
+   matching the `kid` value in the [RFC7517] response
+4. Reconstruct the signature base from the request using the
+   components listed in `Signature-Input` as specified in [RFC9421]
```
glpatcern, line 1504: Also here ideally this can be a single function call that could be suggested as an example.

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
+5. Verify the signature using the appropriate algorithm
+   (e.g., Ed25519 [RFC8032])
```
glpatcern, line 1506: And here too.

But please tell me if you think this is too prescriptive!

glpatcern (commented):
So for the code snippets:

mickenordin:
> So for the code snippets:

There is e.g https://pypi.org/project/http-message-signatures/ but I am not sure that adding snippets using this is really helpful.

This is their example:

```
from http_message_signatures import HTTPMessageSigner, HTTPMessageVerifier, HTTPSignatureKeyResolver, algorithms, http_sfv
import requests, base64, hashlib

class MyHTTPSignatureKeyResolver(HTTPSignatureKeyResolver):
    keys = {"my-key": b"top-secret-key"}

    def resolve_public_key(self, key_id: str):
        return self.keys[key_id]

    def resolve_private_key(self, key_id: str):
        return self.keys[key_id]

request = requests.Request('POST', 'https://example.com/foo?param=Value&Pet=dog', json={"hello": "world"})
request = request.prepare()
request.headers["Content-Digest"] = str(http_sfv.Dictionary({"sha-256": hashlib.sha256(request.body).digest()}))

signer = HTTPMessageSigner(signature_algorithm=algorithms.HMAC_SHA256, key_resolver=MyHTTPSignatureKeyResolver())
signer.sign(request, key_id="my-key", covered_component_ids=("@method", "@authority", "@target-uri", "content-digest"))

verifier = HTTPMessageVerifier(signature_algorithm=algorithms.HMAC_SHA256, key_resolver=MyHTTPSignatureKeyResolver())
verifier.verify(request)
```

Easy enough to look up when implementing, but not really helping you understand what is going on IMO.

glpatcern:
> Easy enough to look up when implementing, but not really helping you understand what is going on IMO.

Well it's not too bad after all, but I take your point that it might be too much to include such code snippets in an Internet Draft. So let's give it try without and we can bring this to the WG anyway. Let me re-review for just a couple of fixes

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
+"@method": POST
+"@target-uri": https://receiver.example.org/ocm/shares
+"content-digest": sha-256=:LkpHyFOVbBDPxc7YbHDOWNzAv88qWuVfLNf4TUf9Uo8=:
+"@signature-params": ("@method" "@target-uri" "content-digest")\
+  ;created=1618884475;keyid="sender.example.org#key1";alg="ed25519"
+```
 
-Here is an example of how to verify the signature using the headers,
-the signature and the public key:
+Sign this base using, for example Ed25519 ([RFC8032]) to produce the
```
```suggestion
Sign this base using for example Ed25519 ([RFC8032]) to produce the
```

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
 ```
-clear = {
-    'content-length': length_of(payload),
-    'date': 'Mon, 08 Jul 2024 14:16:20 GMT',
-    'content-digest': 'SHA-256=' + base64_encode(hash('sha256',
-          utf8_encode(payload))),  # Recompute digest for verification
-    'host': 'sender-fqdn',
-}
+Signature-Input: sig1=("@method" "@target-uri" "content-digest")\
+  ;created=1618884475;keyid="sender.example.org#key1";alg="ed25519"
```
```suggestion
  ;created=<timestamp>;keyid="sender.example.org#key1";alg="ed25519"
```

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
+Signature: sig1=:dGVzdCBzaWduYXR1cmUgdmFsdWU=:
```
```suggestion
Signature: sig1=:<signature-value>=:
```

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
-    signatory,
-  - `headers` is used to generate the clear version of the
-    signature and MUST contain at least `content-length`, `date`,
-    `content-digest` and `host`,
-  - `signature` is the encrypted version of the
-    signature.
+```
+"@method": POST
+"@target-uri": https://receiver.example.org/ocm/shares
+"content-digest": sha-256=:LkpHyFOVbBDPxc7YbHDOWNzAv88qWuVfLNf4TUf9Uo8=:
```
```suggestion
"content-digest": sha-256=:<digest-value>=:
```

IETF-RFC.md
```diff
@@ -1430,126 +1409,95 @@ out of scope for this specification: a mechanism similar to the
...
+"@signature-params": ("@method" "@target-uri" "content-digest")\
+  ;created=1618884475;keyid="sender.example.org#key1";alg="ed25519"
```
```suggestion
  ;created=<timestamp>;keyid="sender.example.org#key1";alg="ed25519"
```

glpatcern (commented):
Minimal fixes - I think it's better to have abstract values (as in RFC7515 actually), especially without any actual code snippet

IETF-RFC.md
glpatcern: (also is the `\` part of the content or is it there for display purposes only?)

mickenordin: The slash is there for display only, either it is there or we exceed 72 chars. No great options there I think :/

mickenordin:
> Minimal fixes - I think it's better to have abstract values (as in RFC7515 actually), especially without any actual code snippet

Added in them now.

IETF-RFC.md
glpatcern: I see. Getting inspiration from RFC7515 I just added a clause "line breaks for display purposes only" and removed the `\`, how does it look like? If it's OK for you, feel free to merge!

glpatcern (approved):

PR #338, opened 2026-02-17
Fix ASCII art diagrams in Appendix D being mangled in text output

mickenordin:
Remove the 'artwork' info string from fenced code blocks so kramdown-rfc produces <artwork> elements instead of <sourcecode> or <spanx>, which collapsed the diagrams into single lines.

claude suggested this and verified with kramdown-rfc tool


glpatcern (approved):
Interestingly Claude suggested (to me) to _add_ that keyword to fix the formatting :-)

PR #348, opened 2026-03-18
Clarify copyright status of specification for DFSG compatibility

petterreinholdtsen:
Adding new section "Copying condition" lifted from <URL: https://wiki.debian.org/NonFreeIETFDocuments > to ensure the specification text follow the Debian Free Software Guidelines and can be distributed as part of programs in Debian outside the non-free set.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -1352,6 +1352,14 @@ signed using HTTP Signatures.  Bearer tokens MUST be treated as
 confidential and never logged, persisted beyond their lifetime, or
 transmitted over unsecured channels.
 
+# Copying conditions
+
+The author(s) agree to grant third parties the irrevocable right to
+copy, use and distribute the work, with or without modification, in
+any medium, without royalty, provided that, unless separate permission
+is granted, redistributed modified works do not contain misleading
+author, version, name of work, or endorsement information.
+
 # References
 
 ## Normative References


glpatcern (approved):
Thanks Petter! This looks fine to me, @lisad @thibmeu how would you consider this paragraph?
For some context, this came out today at the CS3 workshop's [OCM session](https://indico.cern.ch/event/1560960/sessions/628756/#20260318).

cc @mickenordin

mickenordin (commented):
I asked Andy Newton about this, and his reply was that it may be ok to add such a clause, since the authors have copyright and is simply granting the IETF a license. BUT, to be absolutely sure we are not causing an issue here, he said that we should probably run this by ipr@ietf.org, and in addition the templates for ID may not allow for it. So I will do that, and request feedback from the IETF people about this. Personally I see no issue with it, and I am all for granting additional explicitly open source style licenses if someone needs it.

petterreinholdtsen:
Very good to see the positive reception and willingness to find a good
solution.  As a small data point supporting the request, I am happy to
be able to tell you that I have been approached by three CS3 conference
attendees who was very glad I brought up the issue and hoped the
specification would be aligned with the Debian Free Software Guidelines
(which as you might know, was the basis for the Open Source Definition).

Keep up the good work. :)

-- 
Happy hacking
Petter Reinholdtsen

mickenordin:
Apparently I emailed the wrong address, and can't find the right to use. BUT I found this: https://trustee.ietf.org/about/faq/#copyright-licenses


    Under RFC 5378, each Contributor grants the IETF Trust a broad (worldwide, royaltyfree) license to copy, publish, display, translate and distribute his or her IETF Contributions. In addition, unless certain legends are included in a Contribution, the IETF Trust also obtains the right to modify and create derivative works of these Contributions and to grant sublicenses of those rights to others.

    The license to the IETF Trust is non-exclusive. This means that the author can grant a similar license to any other organization or entity without violating the license granted to the IETF Trust, so long as the license to the IETF Trust is not thereby constrained.

To me that pretty clearly states that this is fine, since it is in no way constrains the IETF to do what it want's with the document.

mickenordin (approved):

glpatcern:
Thanks Micke for the investigation! Merging

PR #352, opened 2026-03-26
Drop legacy ocm-provider discovery endpoint

### LINKED ISSUE #349: Drop /ocm-provider
The legacy discovery endpoint is still mandated in the specification, from the discovery algorithm:

    Step 4: If not, try a HTTP GET with `https://<fqdn>/ocm-provider` as the URL instead.

However, all major vendors now support the .well-known endpoint, as can be gleaned by visiting these URLs:

* https://cloud.nextcloud.com/.well-known/ocm

* https://demo.opencloud.eu/.well-known/ocm

* https://demo.owncloud.com/.well-known/ocm

So I think we have an opportunity to clean up the discovery in a nice way, and we should take it.

mickenordin:
Fixes https://github.com/cs3org/OCM-API/issues/349

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+*-design.md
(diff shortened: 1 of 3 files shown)


mickenordin:
The .gitignore can go in as well, convenient to keep design documents around without pushing them to the repo, I cant push directly to develop so lets just add it in now

glpatcern (approved):
Interesting to see you have some `*-design.md` documents for you ;-) (if you wish you could commit them)
Otherwise, it looks good

PR #353, opened 2026-03-26
Add journaling

mickenordin:
This PR introduces the journaling concept as discussed in:

https://github.com/cs3org/OCM-API/discussions/219


mickenordin:
I will close this, I think the possibility to recover and sync is still useful, but I am thinking that calculating a merkle tree from the current state and comparing those is better

PR #354, opened 2026-03-30
Clarify code flow sender and receiver semantics

MahdiBaghbani:
This PR comes from implementation work, not from editorial cleanup.

I have been working on the OCM code flow across three different platforms: Reva, Nextcloud, and OpenCloudMesh Go.

One of them I have been building from scratch, and the other two are established EFSS deployments with real users and compatibility constraints.

While doing that, I kept hitting the same question: when the current text talks about `exchange-token`, `token-exchange`, `must-exchange-token`, and `tokenEndPoint`, which parts mean provider capability, which parts mean receiver policy, and which parts mean per share contract?

My feeling is that the draft already has most of the needed pieces, but it still does not read like one clear end to end model. This PR is my attempt to make that model easier to read without adding new wire fields.

I do not mean this as the final answer for the topic. If this reading feels too strong, too weak, or simply not the one the group wants, I would like to discuss it. The two diagrams below show the gap I kept hitting and the reading this PR is trying to make easier to follow.

```mermaid
flowchart LR
    subgraph A["What kept feeling unclear in practice"]
        A1["exchange-token in discovery"]
        A2["Could be read as sender support"]
        A3["Could be read as receiver support"]
        A4["Could also be read as both"]
        A5["token-exchange in criteria"]
        A6["Easy to read as another capability"]
        A7["tokenEndPoint"]
        A8["Direction not obvious from surrounding text"]
    end

    subgraph B["Reading this PR proposes"]
        B1["exchange-token"]
        B2["Provider supports the code flow"]
        B3["As Sending Server, it hosts tokenEndPoint"]
        B4["As Receiving Server, it can honor strict inbound shares"]
        B5["token-exchange"]
        B6["Receiver side policy for strict inbound shares"]
        B7["must-exchange-token"]
        B8["Per share strict requirement"]
        B9["tokenEndPoint"]
        B10["Hosted by sender and called by receiver"]
    end

    A1 --> A2
    A1 --> A3
    A1 --> A4
    A5 --> A6
    A7 --> A8
    B1 --> B2
    B2 --> B3
    B2 --> B4
    B5 --> B6
    B6 --> B8
    B7 --> B8
    B9 --> B10
```

```mermaid
flowchart TD
    ET["exchange-token in discovery"]
    TX["token-exchange in criteria"]
    EP["tokenEndPoint in discovery"]
    MS["must-exchange-token on the share"]

    ET --> P["One provider-level code-flow capability"]
    P --> S["Sending Server role"]
    P --> R["Receiving Server role"]

    S --> S1["Hosts tokenEndPoint"]
    EP --> S1

    R --> R1["Can honor inbound strict shares"]
    R --> R2["Receiver policy for inbound shares"]
    TX --> R2
    R2 --> R3["If advertised, inbound shares must include must-exchange-token"]
    R3 --> MS

    MS --> M1["Strict share contract"]
    M1 --> M2["Receiver must exchange sharedSecret before access"]
```

```mermaid
flowchart TD
    A["Sending Server wants to create a share"]
    A --> B["Query Receiving Server discovery"]
    B --> C{"Receiver advertises token-exchange"}

    C -- "Yes" --> D{"Sender exposes exchange-token and tokenEndPoint"}
    D -- "Yes" --> E["Include must-exchange-token"]
    E --> F["Create strict share"]
    F --> G["Do not fall back to legacy access for that share"]
    D -- "No" --> H["Do not create the share"]

    C -- "No" --> I{"What is the sender policy for this share"}
    I -- "Strict" --> J["Include must-exchange-token voluntarily"]
    J --> F
    I -- "Legacy" --> K["Send legacy share without must-exchange-token"]
```

```mermaid
flowchart TD
    A["Receiving Server accesses the shared resource"]
    A --> B["Inspect protocol.webdav.requirements"]
    B --> C{"must-exchange-token present"}

    C -- "Yes" --> D["POST sharedSecret to the Sending Server tokenEndPoint"]
    D --> E["Use only the bearer token for WebDAV access"]

    C -- "No" --> F{"Sender discovery exposes exchange-token and tokenEndPoint"}
    F -- "Yes" --> G["Receiver may attempt token exchange"]
    G --> H{"Exchange succeeds"}
    H -- "Yes" --> E
    H -- "No" --> I["Fall back to sharedSecret access"]
    F -- "No" --> I
```

What I am trying to say is pretty simple.

- `exchange-token` stays at provider level. It means the provider supports the code flow. When that provider acts as Sending Server, it hosts `tokenEndPoint`. When it acts as Receiving Server, it can accept inbound shares that require token exchange.
- `token-exchange` stays in discovery `criteria`. I read it as receiver side policy. A provider advertising it is saying that inbound shares must use the code flow.
- `must-exchange-token` stays at share level. It is the strict contract for that specific share.
- `tokenEndPoint` is hosted by the Sending Server and called by the Receiving Server to exchange the `sharedSecret` for a short lived bearer token.

Part of why I think this is worth making more explicit is that current implementations seem to hit the same ambiguity from different angles.

Nextcloud already advertises and uses the relevant code flow pieces, but its outgoing strict share behavior is not yet modeled as an explicit sender side contract.

Reva already has strong receiver side enforcement for strict shares, but its sender entry points still do not consistently emit the strict share requirement.

OpenCloudMesh Go currently has the clearest explicit share level model, but it still relies on local configuration where this PR proposes a clearer discovery driven preflight, and it also keeps an interop leniency escape hatch.

I am not presenting any one of those implementations as the source of truth. I am only using them as a sign that the current wording still leaves room for different reasonable readings.

One design choice worth mentioning is that this PR does not split `exchange-token` into separate sender side and receiver side capabilities.

I did think about that option, because it would make some role differences easier to read when looking only at discovery.

I decided not to do that here because it would add wire level churn and would turn this from a clarification pass into a protocol extension.

The proposal in this PR is narrower: `exchange-token` stays the provider level signal that the provider supports the code flow, while `token-exchange` in `criteria` carries the strict receiver side policy and `must-exchange-token` carries the strict per share contract.

If others think the role split should eventually be expressed through separate discovery capabilities, I am interested in that discussion too. For this PR, I thought it was better to first see whether the current model could be made coherent without adding new fields.

The point where I expect the most debate is probably the sender side preflight rule.

In the reading proposed here, if the Receiving Server advertises `token-exchange` and the Sending Server can provide `exchange-token` plus `tokenEndPoint`, the sender should mark the share as strict with `must-exchange-token`.

If the receiver requires code flow but the sender cannot provide it, the sender should not create the share.

If the receiver does not require it, the sender may still choose to send a strict share.

If people think that is the wrong place to draw the line, or that the optional exchange path for non strict shares should be described differently, I would like to hear that. My main goal with this PR is to make the model easier to discuss and easier to implement in a consistent way.


MahdiBaghbani:
I also admit that I have mistaken the `token-exchange` with `exchange-token` multiple times during  building my OCM mental model or developing 😄

mickenordin (approved):
Very good!

glpatcern (approved):

PR #358, opened 2026-04-05
MLS over OCM Proposal

mickenordin:
This is a proposed architecture for federated groups in OCM leveraging MLS.

### DIFF
--- a/work/mls-over-ocm/cm-mls-federated-groups.md
+++ b/work/mls-over-ocm/cm-mls-federated-groups.md
@@ -1,0 +1,883 @@
+# Federated Groups in OCM using MLS
+
+---
+
+## Abstract
+
+This document proposes an extension to the Open Cloud Mesh (OCM)
+protocol to support federated groups as receiving parties of shares.
+This is achieved using the Messaging Layer Security (MLS) protocol
+[RFC9420] as a group management layer. MLS is used for establishing and
+rotating a shared group key across federated group members, as well as
+for maintaining group state. This gives us not only a way of federating
+group membership, but also a standardized way of distributing encryption
+keys in a cryptographically secure way, so that we can optionally
+encrypt and decrypt files shared across a group. MLS usage in OCM is not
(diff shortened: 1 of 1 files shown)


KrausMatthias:
Sounds great! 

I'm a bit worried about the added complexity, but I guess it might be worth it.

This would also act as a replacement/alternative to sharedSecret or Token Flow? 

Data Encryption support might be advertised via a capability?

mickenordin:
> Sounds great!
> 
> I'm a bit worried about the added complexity, but I guess it might be worth it.

This is an optional part that you have to advetize.

> This would also act as a replacement/alternative to sharedSecret or Token Flow?

The idea is to keep the current OCM messages as is, so not in my proposal.
> Data Encryption support might be advertised via a capability?

You don't need to, since it is you who share encrypted material, you would only do that if you support encryption. Decryption support on the other hand could be made optional and signaled via a capability, but since you have to implement decryption any way, using the same algorithms, I think it should come as a package.

MahdiBaghbani (approved):
I'm very interested in this and as someone who has followed your progress in code (Amity) I have seen a little bit of MLS work in action

Although I'm quite uninformed about this and need to read more to understand it better

I would like to see a name different than DS though because it already means "Directory Service" in OCM, overloading them isn't ideal for me :-)

It's a cool thing to add since we have also work on group sharing or even resharing!

glpatcern (approved):
I see there's a good potential, though I still need to digest all in my brain to see how far we can go vs "is it too complex for OCM?".

Micke, do you plan to present this at the interim? Or shall we have a dedicated follow-up meeting?

Also, IMHO this would make for a separate RFC rather than merging into the main I-D.

mickenordin:
> Micke, do you plan to present this at the interim? Or shall we have a dedicated follow-up meeting?

I wasn't at first, but now I am working on slides anyway :sweat_smile: 

> Also, IMHO this would make for a separate RFC rather than merging into the main I-D.

That is my thinking too

PR #360, opened 2026-04-19
update: notif research and add JSON evidence

MahdiBaghbani:
I tried to manually bring up Nextcloud instances with Spreed (talk) installed and also modified the oCIS Reva and the CERNBox Reva to log the full payload for notifications


glpatcern (approved):
Very good work Mahdi! I'll bring it to the meeting tomorrow :+1:

PR #361, opened 2026-04-20
fix: reva and ocis reva wrong mix, MD layout

MahdiBaghbani:


glpatcern (approved):
As discussed via Matrix, this is an amendment of #360.

PR #362, opened 2026-05-04
Various fixes for http-sig

mickenordin:
I noticed that the alg value was not one of the allowed options. ed25519 is what we use elsewhere, so seems like a better fit than rsa-v1_5-sha256 which is what rsa-sha256 is actually called in https://datatracker.ietf.org/doc/html/rfc9421

As for the JWS alg parameter ES256 is recommended+ in https://www.rfc-editor.org/rfc/rfc7518.html so choosing that.

I also wanted to nail down which parts of the message to cover in signature so that you can not switch the actual payload, and also protect against replay attacks.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -1563,8 +1563,11 @@ breaks in @signature-params for display purposes only):
 <sourcecode type="http">
 "@method": POST
 "@target-uri": https://receiver.example.org/ocm/shares
-"content-digest": sha-256=:[digest-value]=:
-"@signature-params": ("@method" "@target-uri" "content-digest");
+"content-digest": sha-256=:[digest-value]:
+"content-length": [body-length]
+"date": [date]
+"@signature-params": ("@method" "@target-uri" "content-digest"
+    "content-length" "date");
     created=[timestamp];
     keyid="sender.example.org#key1";
     alg="ed25519"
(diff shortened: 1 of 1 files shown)


IETF-RFC.md
```diff
@@ -1563,37 +1566,83 @@ breaks in @signature-params for display purposes only):
...
+headers.
+
+A symmetric signing algorithm MUST NOT be used to sign the
+request, as the Receiving Server would not be able to verify the
+signature without having access to the shared secret in advance.
+
 ## Verifying a Signature (Receiver)
 
+Verifiers MUST locate the ocm-labeled entry and verify only that one,
+if multiple `ocm` signatures are present, the entire message MUST be
```
```suggestion
Verifiers MUST locate the ocm-labeled entry and verify only that one.
If multiple `ocm` signatures are present, the entire message MUST be
```

IETF-RFC.md
```diff
@@ -1563,37 +1566,83 @@ breaks in @signature-params for display purposes only):
...
+reject signatures that omit any of the above components or the `created`
+parameter, and MUST reject signatures whose `created` value is more than
+a small implementation-defined skew tolerance in the future, or older
+than the verifier's freshness window.
+
+A `Content-Digest` header value carrying multiple algorithms MUST have
+every recognised digest match the body; a single match alongside a
+recognised mismatch MUST be treated as an integrity failure.
+
+A request signed in the context of OCM MUST include one and only one,
```
```suggestion
A request signed in the context of OCM MUST include one and only one
```

glpatcern (changes requested):
Very good contribution and reality check Micke. I think it's all fine semantically. Apart the trivial fixes below, I'd ask you to put two spaces after `.` as it's customary with IETF.

MahdiBaghbani:
The `ocm=` part is a nice touch 👍🏼

mickenordin:
> Very good contribution and reality check Micke. I think it's all fine semantically. Apart the trivial fixes below, I'd ask you to put two spaces after `.` as it's customary with IETF.

Yes, of course. Sorry about that.

IETF-RFC.md
```diff
@@ -1563,37 +1566,83 @@ breaks in @signature-params for display purposes only):
...
+components:
+
+  - "@method"             - HTTP method
+  - "@target-uri"         - full request URI (scheme, authority,
+                            path, query)
+  - "content-digest"      - [RFC9530] digest of the body
+  - "content-length"      - bound message size
+  - "date"                - bound clock time
+
+The Signature-Input parameters MUST include `created`. Verifiers MUST
```
```suggestion
The Signature-Input parameters MUST include `created`.  Verifiers MUST
```

glpatcern (commented):
No problem! I'll merge this one myself

glpatcern (approved):

PR #364, opened 2026-05-06
Minor fixes to get all examples uniformly using example.org

glpatcern:


mickenordin (changes requested):
The two changes at the bottom example-1.org and example-2.org would move away from the rfc format, so we should keep the subdomains there example.org or use example.org and example.com

glpatcern:
Right, I was hesitant with that, I like the second better - just not to give the impression that a federation can exist only between subdomains.

mickenordin (approved):
👍

PR #365, opened 2026-05-07
JWT: Change from opaque token to signed JWT

mickenordin:
Define how the token is constructed


MahdiBaghbani (approved):

glpatcern (changes requested):
I'm late here but I just sent a comment on the mailing list. Let's settle on that prior to merging.

glpatcern:
As discussed at the meeting today and in the mailing list, this will become a separate I-D as it does not affect the core OCM protocol.

PR #367, opened 2026-05-25
New edition of the webapp protocol

glpatcern:
Following discussions in the mailing list, this is the current proposal, for now only in terms of OpenAPI - the I-D is to come after we have consensus on this one.

A short summary:
* The Discovery endpoint now includes `-receive` protocols, to signal the ability for an endpoint to receive a share (whereas so far what was advertised is the ability to send and serve shares). This was implicit with `webdav`, but it's not for other optional protocols.
* Specifically for `webapp`, the `webapp-receive` protocol includes a `targets`, to signal all possible targets the endpoint can support.
* The `webdav-uri` capability has been moved into a `uri` attribute of the `webdav-receive` protocol, following the new schema. This was never used so far, so such breaking change is not expected to actually break any implementation.
* The `webapp` share now includes `targets`, `permissions` (replacing `viewMode`), `appName`, and `appIcon`.
* Absolute URIs are not deprecated any longer.

Questions:
* Is there any better idea to represent the `-receive` protocols, whilst preserving backwards compatibility for the `webdav` protocol?
* Does the `webapp` share payload capture all that was discussed so far?

### DIFF
--- a/spec.yaml
+++ b/spec.yaml
@@ -409,27 +409,81 @@ components:
                       The top-level WebDAV path at this endpoint. In order to access
                       a remote shared resource, implementations SHOULD use this path
                       as a prefix (see sharing examples).
-                  webapp:
-                    type: string
+                  webdav-receive:
+                    type: object
                     description: >
-                      The top-level path for web apps at this endpoint. In order to
-                      access a remote web app, implementations SHOULD use this path
-                      as a prefix (see sharing examples).
+                      Implementations that support receiving WebDAV shares SHOULD
+                      advertise them here. Note though that older implementations MAY
+                      not support this property.
+                    properties:
+                      uri:
+                        type: string
+                        description: >
+                          This property signals that this endpoint is capable of
+                          receiving `webdav` shares with the given URI format.
+                          The value MUST be either `"absolute"` or `"relative"`.
+                        enum:
+                          - absolute
+                          - relative
+                  webapp:
+                    type: object
+                    description: >
+                      Implementations that support sending WebApp shares MUST advertise
+                      them here, with an empty object as value.
+                  webapp-receive:
+                    type: object
+                    description: >
+                      Implementations that support receiving WebApp shares MUST
+                      advertise them here.
+                    properties:
+                      targets:
+                        type: array
+                        description: >
+                          The target values this endpoint supports as receiver of
+                          a webapp share, similar to the target attributes in an
+                          HTML `<a>` tag.
+                          - "blank" signals that this endpoint supports opening
+                            the URI in a new window or tab.
+                          - "iframe" signals that this endpoint supports opening
+                            the URI in the same browsing context, within an iframe,
+                            and that it can set CORS headers.
+                          - "popup" signals that this endpoint supports opening
+                            the URI in an embedded popup window, and that it can
+                            set CORS headers.
+                        items:
+                          type: string
+                          enum:
+                            - blank
+                            - iframe
+                            - popup
                   ssh:
                     type: string
                     description: >
                       The top-level address in the form `host:port` of an endpoint
                       that supports ssh and scp with a public/private key based
                       authentication.
-                additionalProperties:
-                    type: string
+                  ssh-receive:
+                    type: object
                     description: >
-                      Any additional protocol supported for this resource type MAY
+                      Implementations that support receiving SSH shares MUST
+                      advertise them here, with an empty object as value.
+                additionalProperties:
+                    type: object
+                    description: >
+                      Any additional protocol supported for this resource type SHOULD
                       be advertised here, where the value MAY correspond to a top-level
-                      URI to be used for that protocol.
+                      URI to be used for that protocol, or any other relevant
+                      attribute required for that protocol. Similarly, additional
+                      receiving capabilities for custom protocols SHOULD be advertised.
                 example:
                   webdav: /remote/dav/ocm/
-                  webapp: /apps/ocm/
+                  webdav-receive: {
+                    "uri": "absolute"
+                  }
+                  webapp: {}
+                  webapp-receive: {
+                    "targets": ["blank", "iframe"]
+                  }
                   talk: /apps/spreed/api/
         capabilities:
           type: array
@@ -446,10 +500,9 @@ components:
               - invite-wayf
               - notifications
               - protocol-object
-              - webdav-uri
           example:
-            - webdav-uri
             - protocol-object
+            - invites
             - http-sig
         criteria:
           type: array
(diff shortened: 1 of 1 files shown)


spec.yaml
```diff
@@ -801,16 +885,20 @@ components:
...
               webapp:
-                uri: 7c084226-d9a1-11e6-bf26-cec0c932ce01
+                uri: https://apps.example.org/codimd/7c084226-d9a1-11e6-bf26-cec0c932ce01
                 sharedSecret: hfiuhworzwnur98d3wjiwhr
-                viewMode: read
+                targets:
+                  - blank
+                permissions:
+                  - read
+                appName: CodiMD
```
mickenordin: Maybe we should add the appIcon example here as well. Should we allow only links for the appIcon, or data uri:s as well?

glpatcern: Good point, done. I've added a generic link to an "asset", I find this is quite typical. Not sure I understand what you mean by data URIs?

mickenordin: This is a data uri:

```
data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAEsAAABDCAYAAADUHfiHAAABg2lDQ1BJQ0MgcHJvZmlsZQAAKJF9kT1Iw0AcxV9TS0UrDnaQ4pChOrWLijiWKhbBQmkrtOpgcukXNDEkKS6OgmvBwY/FqoOLs64OroIg+AHiLjgpukiJ/0sKLWI8OO7Hu3uPu3eA0Koz1exLAKpmGdlUUiwUV8TgKwIIYhARxCRm6uncQh6e4+sePr7exXmW97k/x5BSMhngE4kTTDcs4nXimU1L57xPHGZVSSE+J44ZdEHiR67LLr9xrjgs8Mywkc/OEYeJxUoPyz3MqoZKPE0cVVSN8oWCywrnLc5qvcE69+QvDJW05RzXaY4hhUWkkYEIGQ3UUIeFOK0aKSaytJ/08Eccf4ZcMrlqYOSYxwZUSI4f/A9+d2uWpybdpFASCLzY9sc4ENwF2k3b/j627fYJ4H8GrrSuf6MFzH6S3uxq0SNgeBu4uO5q8h5wuQOMPumSITmSn6ZQLgPvZ/RNRWDkFhhYdXvr7OP0AchTV0s3wMEhMFGh7DWPd/f39vbvmU5/P6hecry8V+VMAAAABmJLR0QA/wD/AP+gvaeTAAAACXBIWXMAAA7DAAAOwwHHb6hkAAAAGXRFWHRDb21tZW50AENyZWF0ZWQgd2l0aCBHSU1QV4EOFwAAAr9JREFUeNrt209Ik2EcB/Dv+/q6rW1pEzf/bkkpWGloEVEQXupQCFIIGWbRIUp2CUK85CG7FR08dIvoIhQdPBTsUHQJIiIRFVcammia+X/NOfdH3w5mznDbc+ndO/x+j+/zwtiH3+/Z733GK6mqqiImyyNDmPV0wd/zEcvDg2DWYy47CCX2wvC9Fsy/9VBm26ibWN7mSwh86adJ3EiQAWD04V1CJS2sNcjBka+YefWCGEkLS4Y89/olIUQry9f9gRCilRWZnyWEYGUpkbnphPe43K0wl5bveCvZbNk6Z22XvPomVtUGGAmIRSxiEYtYxCIBsYiV6ihafMjU86dYDS7rGiLDYoGp0AljgRO79pWlEOvZE0QW59PngCHTgKKrzShovM42THrAEAnj++MOfHY3Eks0S95e9F+p1RhLktIWbGV8FN/ut7GyRDPr6UJ4ekojrK3/46ZlFt69YRuKxt/Xrc3oYK2oQvSXT9cYoclxhGd+xl0PT//QBqu0vUP3lRMYHID35sW465G5GW7wfyd4syXxthuNEosP0sQiFrGIxRCLWMRKfRRljw3RxYW4N/RdPgdDrkPXX8JU5EJJS/v/xzLk2BNihSbGEJoY0zXW6pJfmza0Vh5J//7Q6LxMttWcSX8sjc7L5Kzq47BWVHP3Fv013N/2AMrubLahCJbBkY8DjzphLHSyfETmLJOzBIc7Pdh76w5MxSXcs7YbHf694KhrgKOuAaHJcQSGvAhPTWJ1Jahrq8yc3NRgbcRY6GRb8nGHWMQiFrGIxRCLWMTSW5SlgV6okfCOh/B9ep9wPcNihdRzoUble9LJYz1UBRkSO1EkBnseZKhrlBBI1tETrCzR2GvrWVkicd64/Wd0YGUljO3UaeQ3XFsfHQCVInGSV98El7t1c84CJKrExFJeiexjJ5F79jyMBcVb1n4DUgS4+Ol18CEAAAAASUVORK5CYII=
```
It encodes an image directly in the uri, In this case the SUNET logo

mickenordin: I don't see a reason not to permit it, you can just set `<img src=` to a data uri, just as you would another uri, but the visitors don't need to fetch it separately from the remote.

spec.yaml
```diff
@@ -801,16 +885,21 @@ components:
...
+                uri: https://cloud.example.org/remote/dav/ocm/7c084226-d9a1-11e6-bf26-cec0c932ce01
                 sharedSecret: hfiuhworzwnur98d3wjiwhr
                 permissions:
                   - read
                 requirements:
                   - mfa-enforced
               webapp:
-                uri: 7c084226-d9a1-11e6-bf26-cec0c932ce01
+                uri: https://apps.example.org/codimd/7c084226-d9a1-11e6-bf26-cec0c932ce01
                 sharedSecret: hfiuhworzwnur98d3wjiwhr
```
mickenordin, line 897: In our example here, the sharedSecret is the same for webapp and webdav. I think this is important, and how it must work for integration to be smooth, because that means that the webapp server can use its access token to access any files associated with the share, meaning that both shares are part of a bigger coherent share where both the webdav and webapp go together. I don't think we need to say anything about this in the spec.yaml, but in the I-D it will be important to mention.

spec.yaml
glpatcern: Right, I'd even go further and recommend such a URI!

spec.yaml
glpatcern: Fully agreed. There was already a reference to "multi-protocol shares", this is why we bundle multiple protocols in the same share payload after all.

spec.yaml
glpatcern: Done!

mickenordin:
I think all of this is in: https://github.com/cs3org/OCM-API/pull/368

glpatcern:
OK, let's close this one but we should not squash the commits in #368 so we keep the history

PR #366, opened 2026-05-25
Rename criteria in a more consistent way

glpatcern:
This is a first pass to address #359.

PR #368, opened 2026-06-01
Revamped and enhanced webapp protocol in I-D and spec.yaml

mickenordin:
I also normalized the vocabulary for targets to blank, redirect and iframe, as there were some inconsistencies using popup before. Both permissions and targets are required, so as to make everything explicit.

### DIFF
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -692,25 +692,52 @@ contain the following information about its OCM API:
     ~~~
             {
               "webdav": "/remote/dav/ocm/",
-              "webapp": "/app/ocm/",
+              "webdav-receive": { "uri": "absolute" },
+              "webapp": {},
+              "webapp-receive": { "targets": ["blank", "iframe"] },
               "talk": "/apps/spreed/api/"
             }
     ~~~
     {: type="json"}
 
+    The `protocols` object distinguishes a server's role for each
+    protocol: a property named after the protocol (e.g. `webdav`,
+    `webapp`, `ssh`) advertises support for acting as a Sending
+    Server, while a property suffixed with `-receive` (e.g.
+    `webdav-receive`, `webapp-receive`, `ssh-receive`) advertises
+    support for acting as a Receiving Server.
+
     Fields:
     - webdav (string) - The top-level WebDAV [RFC4918] path at this
       endpoint.  In order to access a Remote Resource, implementations
       SHOULD use this path as a prefix (see sharing examples).
-    - webapp (string) - The top-level path for web apps at this
-      endpoint.  In order to access a remote web app, implementations
-      SHOULD use this path as a prefix (see sharing examples).
+    - webdav-receive (object) - Advertised by implementations that
+      support receiving WebDAV shares.  It contains a `uri` property
+      whose value MUST be either `"absolute"` or `"relative"`,
+      signalling the URI format this endpoint accepts.  Note that
+      older implementations MAY not support this property.
+    - webapp (object) - Advertised, as an empty object, by
+      implementations that support sending WebApp shares.
+    - webapp-receive (object) - Advertised by implementations that
+      support receiving WebApp shares.  It contains a `targets`
+      array listing the ways this endpoint is able to present a
+      WebApp share to the user.  A subset of:
+      - `blank` - the endpoint can open the URI in a new window or
+        tab.
+      - `redirect` - the endpoint can navigate the browser to the
+        URI, replacing the current page.
+      - `iframe` - the endpoint can embed the URI in an iframe
+        within its own UI, and can set CORS headers.
     - ssh (string) - The top-level address in the form `host:port`
       of an endpoint that supports ssh and scp with a public/private
       key based authentication.
+    - ssh-receive (object) - Advertised, as an empty object, by
+      implementations that support receiving SSH shares.
     - Any additional protocol supported for this Resource type MAY be
       advertised here, where the value MAY correspond to
-      a top-level URI to be used for that protocol.
+      a top-level URI to be used for that protocol.  Similarly,
+      additional receiving capabilities for custom protocols SHOULD
+      be advertised using a `-receive` suffixed property.
 * OPTIONAL: capabilities (array of string) - The optional capabilities
   supported by this OCM Server.
   As implementations MUST accept Share Creation Notifications
@@ -947,20 +974,49 @@ voluntarily.
     especially in case of `datatx` access type.
 * Protocol details for `webapp` MAY contain:
   - REQUIRED uri (string)
-    A URI to a client-browsable view of the Shared
-    Resource, such that users MAY use the web
-    applications available at the site.  The URI SHOULD
-    be relative, in which case the prefix exposed by
-    the `/.well-known/ocm` endpoint MUST be used.
-    Absolute URIs are deprecated.
-  - REQUIRED viewMode (string)
-    The permissions granted to the sharee.  A subset of:
+    A URI to a client-browsable view of the Shared Resource, such
+    that users MAY use a web application available at the Sending
+    Server.  The URI MUST be absolute, including a hostname.  In
+    case the underlying Resource is a folder, the URI MUST act as a
+    root path, such that files located within the folder are made
+    accessible in the web app by appending their relative path to
+    the URI.
+  - REQUIRED targets (array of strings) - How the recipient SHOULD
+    present the URI to the user.  MUST NOT be empty.  A subset of:
+    - `blank` signals the recipient to open the URI in a new window
+      or tab.
+    - `redirect` signals the recipient to navigate the browser to
+      the URI, replacing the current page.
+    - `iframe` signals the recipient to embed the URI in an iframe
+      within its own UI.  CORS headers MUST be properly set.
+    A Sending Server MUST NOT offer a target that the recipient did
+    not advertise in its `webapp-receive` discovery property.
+  - REQUIRED permissions (array of strings) -
+    The permissions granted to the sharee.  MUST NOT be empty.
+    A subset of:
     - `view` allows access to the web app in view-only mode.
     - `read` allows read and download access via the web app.
     - `write` allows full editing rights via the web app.
+    - `share` allows re-share rights on the Resource.  This only
+      applies to web apps that provide a mechanism for re-sharing.
   - OPTIONAL sharedSecret (string)
-    An optional secret to be used to access the remote
-    web app, for example in the form of a bearer token.
+    A secret for accessing the remote web app, such as a bearer
+    token.  To give access to the remote app, the receiver MUST
+    perform an HTTP POST request to the given `uri`, passing the
+    shared secret in a form field named `access_token` (see
+    [Resource Access](#resource-access)).  To prevent leaking it in
+    logs it MUST NOT appear in any URI.  In a multi-protocol Share
+    that also offers WebDAV, the access requirements provided in the
+    `webdav` part (such as `must-exchange-token`) MUST apply to
+    `webapp` accesses as well.
+  - OPTIONAL appName (string)
+    A human-friendly name of the web application, to be used in user
+    interfaces when referring to this Share.
+  - OPTIONAL appIcon (string)
+    A URI to an icon representing the web application, to be used in
+    user interfaces when referring to this Share.  An embedded data
+    URI MAY be used; alternatively, if a regular URI is used, it MUST
+    be absolute, including a hostname.
 * Protocol details for `ssh` MAY contain:
   - OPTIONAL accessTypes (array of strings) - The type of access
     being granted to the remote resource.  If omitted, it defaults to
@@ -1163,6 +1219,22 @@ protocol required for access.  The procedure is as follows:
    removed in a future release of the Protocol.  If a secret cannot be
    identified (e.g. because `protocol.options` is undefined), then
    the receiver SHOULD discard the share as invalid.
+6. For the specific case where `protocol.webapp` is available and the
+   receiver wants to use it, the receiver MUST present the web app to
+   the user by opening `protocol.webapp.uri` using one of the targets
+   listed in `protocol.webapp.targets` (defaulting to `blank`), chosen
+   among those the receiver advertises in its `webapp-receive`
+   discovery property.  If a `protocol.webapp.sharedSecret` is present,
+   the receiver MUST NOT place it in the URI; instead it MUST deliver
+   it to the web app via an HTTP POST to `protocol.webapp.uri` with the
+   secret carried in a form field named `access_token`.  This is
+   typically achieved with an auto-submitting HTML form whose `target`
+   attribute selects the chosen presentation (e.g. an iframe name,
+   `_blank`, or `_top`).  When the Share also requires token exchange
+   (see step 3.1, applied via the `webdav` part of a multi-protocol
+   Share), the receiver MUST first exchange the `sharedSecret` at the
+   Sending Server's {tokenEndPoint} and POST the resulting bearer token
+   as the `access_token` value.
 
 In all cases, in case the Shared Resource is a folder and the Receiving
 Server accesses a Resource within that shared folder, it SHOULD append
@@ -1928,8 +2000,11 @@ that section.
 |   Protocols      |
 +------------------+
 | - ssh            |
+| - ssh-receive    |
 | - webapp         |
+| - webapp-receive |
 | - webdav         |
+| - webdav-receive |
 | - ...            |
 +------------------+
 ~~~
--- a/IETF-RFC.md
+++ b/IETF-RFC.md
@@ -722,8 +722,9 @@ contain the following information about its OCM API:
       support receiving WebApp shares.  It contains a `targets`
       array listing the ways this endpoint is able to present a
       WebApp share to the user.  A subset of:
-      - `blank` - the endpoint can open the URI in a new window or
-        tab, or do a full page redirect.
+      - `blank` - the endpoint can open the URI in a top-level
+        browsing context, such as a new window or tab, or a full page
+        navigation in the current window.
       - `iframe` - the endpoint can embed the URI in an iframe
         within its own UI, when the Sending Server allows framing
         by this receiver.
@@ -778,7 +779,8 @@ contain the following information about its OCM API:
   - `"must-exchange-token"` - to indicate that when this OCM Server
   acts as Receiving Server, it requires the code flow for all inbound
   shares.  Shares that do not include `must-exchange-token` in
-  their `protocol.webdav.requirements` will be rejected.  An
+  the requirements of each protocol offered for access will be
+  rejected.  An
   OCM Server advertising this criterium MUST also expose the
   `exchange-token` capability.  See the [Code Flow](#code-flow)
   section.
@@ -823,8 +825,8 @@ the Receiving Server's OCM API Discovery endpoint.  If the Receiving
 Server advertises `must-exchange-token` in its `criteria` and the
 Sending Server exposes the `exchange-token` capability with a
 `tokenEndPoint`, the Sending Server MUST include `must-exchange-token`
-in `protocol.webdav.requirements` and MUST NOT fall back to legacy
-shared-secret access.  If the Receiving Server advertises
+in the requirements of each protocol offered for access and MUST NOT
+fall back to legacy shared-secret access.  If the Receiving Server advertises
 `must-exchange-token` but the Sending Server does not expose the
 `exchange-token` capability or does not have a `tokenEndPoint`, the
 Sending Server MUST NOT create the share, as the Receiving Server
@@ -832,6 +834,13 @@ would reject any notification that lacks the code-flow requirement.
 If the Receiving Server does not advertise `must-exchange-token` in its
 `criteria`, the Sending Server MAY still include `must-exchange-token`
 voluntarily.
+
+When the notification includes `protocol.webapp`, the Sending Server
+MUST expose the `exchange-token` capability and a `tokenEndPoint`,
+because WebApp access requires the Receiving Server to exchange
+`protocol.webapp.sharedSecret` before presenting the WebApp to the
+browser.  If the Sending Server cannot offer this code flow, it MUST NOT
+include `protocol.webapp` in the notification.
 
 ## Fields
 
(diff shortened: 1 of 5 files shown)


spec.yaml
```diff
@@ -409,27 +409,80 @@ components:
...
-                    type: string
+                    type: object
                     description: >
-                      Any additional protocol supported for this resource type MAY
+                      Any additional protocol supported for this resource type SHOULD
                       be advertised here, where the value MAY correspond to a top-level
-                      URI to be used for that protocol.
+                      URI to be used for that protocol, or any other relevant
+                      attribute required for that protocol. Similarly, additional
+                      receiving capabilities for custom protocols SHOULD be advertised.
```
MahdiBaghbani, line 482: I think there is a small schema/text mismatch here.

The schema now says every additional protocol value is an object, but the description still says the value may be a top-level URI, and the example below still uses `talk: /apps/spreed/api/`.

If URI strings are still valid for custom protocols, maybe this could allow both shapes?

```suggestion
                additionalProperties:
                    oneOf:
                      - type: string
                      - type: object
                    description: >
                      Any additional protocol supported for this resource type SHOULD
                      be advertised here, where the value MAY correspond to a top-level
                      URI to be used for that protocol, or any other relevant
                      attribute required for that protocol. Similarly, additional
                      receiving capabilities for custom protocols SHOULD be advertised.
```

spec.yaml
```diff
@@ -409,27 +409,80 @@ components:
                       The top-level WebDAV path at this endpoint. In order to access
                       a remote shared resource, implementations SHOULD use this path
                       as a prefix (see sharing examples).
+                  webdav-receive:
+                    type: object
+                    description: >
+                      Implementations that support receiving WebDAV shares SHOULD
+                      advertise them here. Note though that older implementations MAY
+                      not support this property.
+                    properties:
```
MahdiBaghbani, line 420: Could we make `uri` required here?

Without it, `webdav-receive: {}` is valid according to the schema, but I think senders would not know whether to use an absolute or relative URI.

```suggestion
                  webdav-receive:
                    type: object
                    required:
                      - uri
                    description: >
                      Implementations that support receiving WebDAV shares SHOULD
                      advertise them here. Note though that older implementations MAY
                      not support this property.
                    properties:
```

spec.yaml
```diff
@@ -409,27 +409,80 @@ components:
...
+                      Implementations that support sending WebApp shares MUST advertise
+                      them here, with an empty object as value.
+                  webapp-receive:
+                    type: object
+                    description: >
+                      Implementations that support receiving WebApp shares MUST
+                      advertise them here.
+                    properties:
+                      targets:
+                        type: array
```
MahdiBaghbani, line 444: Same idea here: if `webapp-receive` is present, I think `targets` should be required and non-empty.

Otherwise a receiver can advertise WebApp receive support without any usable presentation mode.

```suggestion
                  webapp-receive:
                    type: object
                    required:
                      - targets
                    description: >
                      Implementations that support receiving WebApp shares MUST
                      advertise them here.
                    properties:
                      targets:
                        type: array
                        minItems: 1
```

spec.yaml
```diff
@@ -409,27 +409,80 @@ components:
...
+                          The ways this endpoint is able to present a webapp
+                          share to the user when acting as receiver.
+                          - "blank" signals that this endpoint supports opening
+                            the URI in a new window or tab.
+                          - "redirect" signals that this endpoint supports
+                            navigating the browser to the URI, replacing the
+                            current page.
+                          - "iframe" signals that this endpoint supports
+                            embedding the URI in an iframe within its own UI,
+                            and that it can set CORS headers.
```
MahdiBaghbani: I think the iframe wording could be a bit more precise.

For iframe embedding, the important part is usually that the Sending Server allows the receiver origin to frame the app, for example with `Content-Security-Policy: frame-ancestors`. The receiver cannot set CORS headers for the remote app response.

Maybe something like this?

```suggestion
                          - "iframe" signals that this endpoint supports
                            embedding the URI in an iframe within its own UI,
                            when the Sending Server allows framing by this
                            receiver.
```

spec.yaml
```diff
@@ -720,38 +772,76 @@ components:
...
-                viewMode:
-                  type: string
-                  description: |
-                    The permissions granted to the sharee.
+                    The URI MUST be absolute, including a hostname. In case the
+                    underlying resource is a folder, the URI MUST act as a root path,
+                    such that files located within the folder are made accessible in
+                    the web app by appending their relative path to the URI.
+                targets:
+                  type: array
```
MahdiBaghbani, line 799: The prose says this array is required and must not be empty, but the schema still accepts `targets: []`.

Could you please add `minItems: 1` here?

```suggestion
                targets:
                  type: array
                  minItems: 1
```

spec.yaml
```diff
@@ -720,38 +772,76 @@ components:
...
+                    - "iframe" signals the recipient to embed the URI in an iframe within
+                      its own UI. CORS headers MUST be properly set.
+                  items:
+                    type: string
+                    enum:
+                      - blank
+                      - redirect
+                      - iframe
+                permissions:
+                  type: array
```
MahdiBaghbani, line 816: Same for `permissions`: the text says it must not be empty, so I think the schema should enforce that too.

```suggestion
                permissions:
                  type: array
                  minItems: 1
```

spec.yaml
```diff
@@ -720,38 +772,76 @@ components:
...
-                    in any URI.  In a multi-protocol share scenario with WebDAV, the
-                    access requirements provided in the `webdav` part MUST apply for
-                    `webapp` accesses as well.
+                    A secret for accessing the remote web app, such as a bearer token.
+                    To give access to the remote app, the receiver MUST perform a HTTP
+                    POST request to the given URI, with the shared secret in a form field
+                    named `access_token`. To prevent leaking it in logs it MUST NOT appear
+                    in any URI. In a multi-protocol share scenario with WebDAV, the access
+                    requirements provided in the `webdav` part MUST apply for `webapp`
+                    accesses as well.
```
MahdiBaghbani: I am not fully sure if this is intended, but this wording seems to allow the raw `sharedSecret` to be sent through the browser as `access_token` unless the WebDAV part requires token exchange.

My preference would be to keep `sharedSecret` server-to-server and only post the exchanged bearer token to the WebApp.

```suggestion
                    A secret for accessing the remote web app. To give access to the
                    remote app, the receiver MUST first exchange this value at the
                    Sending Server's `tokenEndPoint` using the Code Flow, then perform
                    an HTTP POST request to the given URI with the resulting bearer
                    token in a form field named `access_token`. The shared secret MUST
                    NOT be exposed to the browser and MUST NOT appear in any URI. In a
                    multi-protocol share scenario with WebDAV, the access requirements
                    provided in the `webdav` part MUST apply for `webapp` accesses as
                    well.
```

spec.yaml
```diff
@@ -801,16 +891,21 @@ components:
               name: multi
               webdav:
                 accessTypes: ['remote', 'datatx']
-                uri: 7c084226-d9a1-11e6-bf26-cec0c932ce01
+                uri: https://cloud.example.org/remote/dav/ocm/7c084226-d9a1-11e6-bf26-cec0c932ce01
                 sharedSecret: hfiuhworzwnur98d3wjiwhr
                 permissions:
                   - read
                 requirements:
                   - mfa-enforced
```
MahdiBaghbani: Small one: I think this example should use the enum value `must-use-mfa`.

```suggestion
                  - must-use-mfa
```

IETF-RFC.md
```diff
@@ -692,25 +692,52 @@ contain the following information about its OCM API:
...
+    - webapp-receive (object) - Advertised by implementations that
+      support receiving WebApp shares.  It contains a `targets`
+      array listing the ways this endpoint is able to present a
+      WebApp share to the user.  A subset of:
+      - `blank` - the endpoint can open the URI in a new window or
+        tab.
+      - `redirect` - the endpoint can navigate the browser to the
+        URI, replacing the current page.
+      - `iframe` - the endpoint can embed the URI in an iframe
+        within its own UI, and can set CORS headers.
```
MahdiBaghbani: Same iframe wording as in the OpenAPI file: I think this should talk about the sender allowing framing, not the receiver setting CORS headers.

```suggestion
      - `iframe` - the endpoint can embed the URI in an iframe
        within its own UI when the Sending Server allows framing by
        this receiver.
```

IETF-RFC.md
```diff
@@ -947,20 +974,49 @@ voluntarily.
...
-    The permissions granted to the sharee.  A subset of:
+    A URI to a client-browsable view of the Shared Resource, such
+    that users MAY use a web application available at the Sending
+    Server.  The URI MUST be absolute, including a hostname.  In
+    case the underlying Resource is a folder, the URI MUST act as a
+    root path, such that files located within the folder are made
+    accessible in the web app by appending their relative path to
+    the URI.
+  - REQUIRED targets (array of strings) - How the recipient SHOULD
+    present the URI to the user.  MUST NOT be empty.  A subset of:
```
MahdiBaghbani: Since `targets` is now required, I think the later `defaulting to blank` wording becomes a bit confusing.

Could we make the non-empty requirement explicit here and handle missing/empty targets as invalid?

```suggestion
  - REQUIRED targets (array of strings) - How the recipient SHOULD
    present the URI to the user.  The `targets` array MUST NOT be
    empty.  A subset of:
```

IETF-RFC.md
```diff
@@ -947,20 +974,49 @@ voluntarily.
...
+    - `blank` signals the recipient to open the URI in a new window
+      or tab.
+    - `redirect` signals the recipient to navigate the browser to
+      the URI, replacing the current page.
+    - `iframe` signals the recipient to embed the URI in an iframe
+      within its own UI.  CORS headers MUST be properly set.
```
MahdiBaghbani: Same iframe point here too: I think this should use framing-policy language instead of CORS.

```suggestion
    - `iframe` signals the recipient to embed the URI in an iframe
      within its own UI when the Sending Server allows framing by this
      receiver.
```

IETF-RFC.md
```diff
@@ -947,20 +974,49 @@ voluntarily.
...
-    web app, for example in the form of a bearer token.
+    A secret for accessing the remote web app, such as a bearer
+    token.  To give access to the remote app, the receiver MUST
+    perform an HTTP POST request to the given `uri`, passing the
+    shared secret in a form field named `access_token` (see
+    [Resource Access](#resource-access)).  To prevent leaking it in
+    logs it MUST NOT appear in any URI.  In a multi-protocol Share
+    that also offers WebDAV, the access requirements provided in the
+    `webdav` part (such as `must-exchange-token`) MUST apply to
+    `webapp` accesses as well.
```
MahdiBaghbani: Similar to the OpenAPI comment: I think the safer contract is that `sharedSecret` is exchanged server-to-server first, and only the resulting bearer token is posted to the WebApp.

If direct POST of the raw `sharedSecret` is intentional, then I think it needs a bit more security text around lifetime/replay/logging.

```suggestion
    A secret for accessing the remote web app.  To give access to the
    remote app, the receiver MUST first exchange this value at the
    Sending Server's {tokenEndPoint} using the Code Flow, then perform
    an HTTP POST request to the given `uri` with the resulting bearer
    token in a form field named `access_token`.  The shared secret MUST
    NOT be exposed to the browser and MUST NOT appear in any URI.  In a
    multi-protocol Share that also offers WebDAV, the access
    requirements provided in the `webdav` part (such as
    `must-exchange-token`) MUST apply to `webapp` accesses as well.
```

IETF-RFC.md
```diff
@@ -947,20 +974,49 @@ voluntarily.
...
+  - OPTIONAL appName (string)
+    A human-friendly name of the web application, to be used in user
+    interfaces when referring to this Share.
+  - OPTIONAL appIcon (string)
+    A URI to an icon representing the web application, to be used in
+    user interfaces when referring to this Share.  An embedded data
+    URI MAY be used; alternatively, if a regular URI is used, it MUST
+    be absolute, including a hostname.
```
MahdiBaghbani: This may be worth constraining a little because receiver UIs will likely render this value directly.

Could we say that data URIs must still be image resources and should be rendered only in an inert image context?

```suggestion
    A URI to an icon representing the web application, to be used in
    user interfaces when referring to this Share.  An embedded data
    URI MAY be used if it identifies an image resource; alternatively,
    if a regular URI is used, it MUST be absolute, including a
    hostname.  Receiving Servers MUST render the icon only in an inert
    image context and MAY reject unsupported or unsafe image types.
```

IETF-RFC.md
```diff
@@ -1163,6 +1219,22 @@ protocol required for access.  The procedure is as follows:
...
+   the receiver MUST NOT place it in the URI; instead it MUST deliver
+   it to the web app via an HTTP POST to `protocol.webapp.uri` with the
+   secret carried in a form field named `access_token`.  This is
+   typically achieved with an auto-submitting HTML form whose `target`
+   attribute selects the chosen presentation (e.g. an iframe name,
+   `_blank`, or `_top`).  When the Share also requires token exchange
+   (see step 3.1, applied via the `webdav` part of a multi-protocol
+   Share), the receiver MUST first exchange the `sharedSecret` at the
+   Sending Server's {tokenEndPoint} and POST the resulting bearer token
+   as the `access_token` value.
```
MahdiBaghbani: Could we avoid the default here and define the no-common-target case?

That would make the receive-side behavior deterministic when the sender's `targets` and the receiver's `webapp-receive.targets` do not overlap. This also keeps the raw `sharedSecret` out of the browser.

```suggestion
6. For the specific case where `protocol.webapp` is available and the
   receiver wants to use it, the receiver MUST present the web app to
   the user by opening `protocol.webapp.uri` using a target selected
   from the intersection of `protocol.webapp.targets` and the targets
   advertised in the receiver's `webapp-receive` discovery property.
   If this intersection is empty, the receiver MUST treat the `webapp`
   option as unusable for this Share.  If a
   `protocol.webapp.sharedSecret` is present, the receiver MUST NOT
   place it in the URI and MUST NOT expose it to the browser.  Instead,
   the receiver MUST first exchange it at the Sending Server's
   {tokenEndPoint} using the Code Flow, then deliver the resulting
   bearer token to the web app via an HTTP POST to
   `protocol.webapp.uri` with the token carried in a form field named
   `access_token`.  This is typically achieved with an auto-submitting
   HTML form whose `target` attribute selects the chosen presentation
   (e.g. an iframe name, `_blank`, or `_top`).
```

MahdiBaghbani (changes requested):
Thanks for working on this, I think the direction is really good.

The split between `targets` and `permissions` makes the WebApp part much easier to reason about, and I also like moving receive-side support into `webapp-receive.targets` instead of overloading the old `viewMode` field.

I left a few comments where I think the text, OpenAPI schema, and JSON schema do not fully agree yet. Most of these look mechanical, but I think they matter because this PR changes the wire contract for discovery and WebApp shares.

The main thing I would like to see before merge is that `schemas/ocm-discovery.json` is updated too. Right now it still describes `webapp` as a string path and does not include `webdav-receive`, `webapp-receive`, or `ssh-receive`, so implementers may get two different discovery contracts from the same repository.

A possible direction for the relevant JSON schema parts would be:

```json
{
  "capabilities": {
    "type": "array",
    "description": "Capability values of 'enforce-mfa', 'exchange-token', 'http-sig', 'invites', 'invite-wayf', 'notifications', and 'protocol-object' are defined in the draft",
    "items": {
      "type": "string"
    }
  }
}
```

And for the `$defs` section:

```json
{
  "resourceType": {
    "type": "object",
    "properties": {
      "name": { "type": "string" },
      "shareTypes": { "type": "array" },
      "protocols": { "$ref": "#/$defs/protocols" }
    },
    "required": ["name", "shareTypes", "protocols"]
  },
  "protocols": {
    "type": "object",
    "minProperties": 1,
    "description": "Additional protocols besides 'webdav', 'webapp', 'datatx', and 'ssh' may be defined.",
    "properties": {
      "webdav": { "type": "string", "pattern": "^/" },
      "webdav-receive": {
        "type": "object",
        "required": ["uri"],
        "properties": {
          "uri": { "type": "string", "enum": ["absolute", "relative"] }
        },
        "additionalProperties": false
      },
      "webapp": { "type": "object", "additionalProperties": false },
      "webapp-receive": {
        "type": "object",
        "required": ["targets"],
        "properties": {
          "targets": {
            "type": "array",
            "minItems": 1,
            "uniqueItems": true,
            "items": { "type": "string", "enum": ["blank", "redirect", "iframe"] }
          }
        },
        "additionalProperties": false
      },
      "datatx": { "type": "string", "pattern": "^/" },
      "ssh": { "type": "string" },
      "ssh-receive": { "type": "object", "additionalProperties": false }
    },
    "additionalProperties": {
      "oneOf": [
        { "type": "string" },
        { "type": "object" }
      ]
    }
  }
}
```

If the intention is instead that custom protocol values must become objects too, then I think the `talk` example and I-D prose should be changed in the same PR.

A few other things I noticed:

- `webdav-uri` seems removed from OpenAPI but still present in the I-D text
- the WebDAV URI wording is different between `spec.yaml` and `IETF-RFC.md`
- WebApp `sharedSecret` handling probably needs one more pass so the raw secret is not accidentally exposed to the browser
- the iframe wording probably wants frame-policy language rather than CORS
- a few arrays are described as required/non-empty, but the schema does not enforce that yet

Could you please have another look at these before merge?

Some of my suggestions are a bit more normative than editorial, so if I misunderstood the intended model, feel free to push back and we can narrow it down.

IETF-RFC.md
MahdiBaghbani: I feel like this request is bit of a nitpick since I have not read the full webapp implementation you did in Nextcloud, so I might have completely got it wrong.

spec.yaml
MahdiBaghbani: I feel like since I haven't worked on the webapps this comment would be wrong in some sense since the webapp is supposed to open in user browser, so it makes sense that the secret is somewhat sent from user browser.

IETF-RFC.md
mickenordin: No, indeed, this is how it is being implemented in Nextcloud, and I think you reading is correct here.

mickenordin:
Thanks for the review @MahdiBaghbani! I think I addressed all your concerns now.

IETF-RFC.md
```diff
@@ -1427,7 +1505,7 @@ Values" registry (using the template from [RFC9553]).
 There are several areas that are not covered by this specification.
 Most importantly we do not provide a way of establishing trust between
 servers, even though some features of the protocol rely on trust, such
-as the `mfa-enforced` requirement.
```
glpatcern: This was addressed by https://github.com/cs3org/OCM-API/pull/366, can we review/merge that one first?

glpatcern (changes requested):
This looks very good. I have a couple of general comments:
 
- for the targets: in my understanding a `popup` would be an embedding similar to `iframe`: are we collapsing the two as technically it's the same? Otherwise, `redirect` may be useful, though the name "redirect" didn't tell me much until I saw the definition. The concern I have is that the receiver may have its own policy to serve a `blank` as a new page or as navigating to the page (including giving both options, like we do in CERNBox), and technically they are the same thing - a bit like an embedded popup and an iframe are different presentations of the same technology.

- I'm totally in for making webapps use the token exchange, so that the sharedSecret never gets shipped to a browser. This now becomes an implicit `requirement`: are we happy with that? How do we phrase it?

IETF-RFC.md
```diff
@@ -947,20 +973,62 @@ voluntarily.
...
-    An optional secret to be used to access the remote
-    web app, for example in the form of a bearer token.
+    A secret for accessing the remote web app.  To give access to the
+    remote app, the receiver MUST first exchange this value at the
+    Sending Server's {tokenEndPoint} using the Code Flow, then perform
+    an HTTP POST request to the given `uri` with the resulting bearer
+    token in a form field named `access_token` (see
+    [Resource Access](#resource-access)).  The shared secret MUST NOT
+    be exposed to the browser and MUST NOT appear in any URI.  In a
+    multi-protocol Share that also offers WebDAV, the access
```
glpatcern: Maybe we can now drop this sentence, the above part is stronger.

spec.yaml
```diff
@@ -801,16 +902,21 @@ components:
               name: multi
               webdav:
                 accessTypes: ['remote', 'datatx']
-                uri: 7c084226-d9a1-11e6-bf26-cec0c932ce01
+                uri: https://cloud.example.org/remote/dav/ocm/7c084226-d9a1-11e6-bf26-cec0c932ce01
                 sharedSecret: hfiuhworzwnur98d3wjiwhr
                 permissions:
                   - read
                 requirements:
                   - must-use-mfa
```
glpatcern, line 950:
```suggestion
                  - must-use-mfa
                  - must-exchange-tokem                 
```

spec.yaml
```diff
@@ -801,16 +902,21 @@ components:
...
-                uri: 7c084226-d9a1-11e6-bf26-cec0c932ce01
+                uri: https://apps.example.org/codimd/7c084226-d9a1-11e6-bf26-cec0c932ce01
                 sharedSecret: hfiuhworzwnur98d3wjiwhr
-                viewMode: read
+                targets:
+                  - blank
+                permissions:
+                  - read
+                appName: CodiMD
+                appIcon: https://apps.example.org/assets/codimd-icon.png
```
```suggestion
                uri: https://apps.example.org/hedgedoc/7c084226-d9a1-11e6-bf26-cec0c932ce01
                sharedSecret: hfiuhworzwnur98d3wjiwhr
                targets:
                  - blank
                permissions:
                  - read
                requirements:
                  - must-exchange-token
                appName: HedgeDoc
                mediaTypes:
                  - text/markdown
                  - text/html
```

glpatcern (changes requested):
I think we're pretty much done, just a couple of comments and suggestions:

spec.yaml
glpatcern: (Just saw this was updated - the above comes from an unmerged commit from the previous PR, let me add it again)

spec.yaml
```diff
@@ -801,16 +929,25 @@ components:
...
                 sharedSecret: hfiuhworzwnur98d3wjiwhr
-                viewMode: read
+                targets:
+                  - blank
+                permissions:
+                  - read
+                requirements:
+                  - must-use-mfa
+                  - must-exchange-token
+                appName: CodiMD
```
```suggestion
                uri: https://apps.example.org/hedgedoc/7c084226-d9a1-11e6-bf26-cec0c932ce01
                sharedSecret: hfiuhworzwnur98d3wjiwhr
                targets:
                  - blank
                permissions:
                  - read
                requirements:
                  - must-use-mfa
                  - must-exchange-token
                appName: HedgeDoc
```

mickenordin:
I think all the feedback is addressed now.

glpatcern:
> I think all the feedback is addressed now.

Yes for the I-D, whereas for the `spec.yaml` somehow my last comment was superseded by a parallel force-push. Can you have a look? I think there's now a duplicate requirement (we both added it...), and the last commit from #367 didn't make it here.

glpatcern (approved):
Now it's all good!

MahdiBaghbani (approved):
I made some edits and I think it's looking good now

glpatcern (approved):
Thanks Mahdi, the extra review was totally needed! Let's go for now and I don't exclude a further review and refinement.

PR #370, opened 2026-06-11
OCM-IP: First draft of the I-D document.

mickenordin:
This PR adds the first draft of the Open Cloud Mesh Integration Protocol, and acknowledges Enrique's contribution in the main OCM spec.

I don't think it will be valuable to work too much on the text of this spec in this pr, rather focusing on IETF ID nits and such (if there are any), the IETF call for adoption procedure will safeguard that we do not introduce a bad document, and obviously we will have to evolve this document incrementally in pr's going forward.

I would appreciate assistance with the CI for building the xml document on pr.


glpatcern:
I'm going through the doc, at least as a first pass. 

I can also help with the CI, I have something already, will commit in this branch.

MahdiBaghbani:
I'm reading this as well I have some points that I will write in coming days in review

mickenordin:
> I'm reading this as well I have some points that I will write in coming days in review

If possible, I think it would be good to invert this flow. I.e. merge this PR, and them work on the content of the documents by making pr:s against it. Remeber that merging this PR makes no difference of the status of the document at the IETF, the call for adoption is what makes it an "official" document in the WG, and I think addressing all concerns as reviews against this PR will be much harder than the opposite, changing the document incrementally after it has been merged.

MahdiBaghbani (dismissed):

IETF-OCM-IP.md
```diff
@@ -0,0 +1,1553 @@
...
+needs no per-share state and no inbound API at all; and an introspected
+mode, in which the Protocol Server validates presented credentials
+through a token introspection endpoint, restoring compatibility with
+Receiving Servers that do not support token exchange.
+
+OCM-IP is a protocol between the Sending OCM Server and its Protocol
+Servers only.  The Receiving Server is not involved in, and does not
+need to be aware of, this protocol: everything it observes is
+indistinguishable from the Sending Server serving the access protocols
+itself.
```
glpatcern:
```suggestion
itself.  For this reason, an OCM Sending Server MAY adopt a
different strategy to interoperate with Protocol Servers, including
e.g. establishing trust via shared keys, without compromising
compliance with the OCM protocol.
```

glpatcern (commented):
I think this is overall OK, let's indeed merge the PR and then we go through the details as part of the adoption process.

I would only explicitly add a paragraph in the introduction:

glpatcern (approved):

PR #371, opened 2026-06-11
Removed the already deprecated draft-cavage pubkey

glpatcern:
This is a "maintenance" PR: as in Nextcloud we now have a RFC-compliant implementation, there's no need to carry over the deprecated draft-cavage public key. This removes it completely.


mickenordin (approved):
🎉

PR #372, opened 2026-06-12
OCM-MLS: I-D for MLS over OCM

mickenordin:
This is the initial draft for Open Cloud Mesh Federated Groups over Message Layer Security (OCM-MLS).


IETF-OCM-MLS.md
```diff
@@ -0,0 +1,1914 @@
...
+
+# References
+
+## Normative References
+
+[OCM] Lo Presti, G., de Jong, M.B., Baghbani, M. and Nordin, M.  "[Open
+Cloud
+Mesh](https://datatracker.ietf.org/doc/draft-ietf-ocm-open-cloud-mesh/)",
+Work in Progress, Internet-Draft, draft-ietf-ocm-open-cloud-mesh-04,
+March 2026.
```
glpatcern: I'd not include the version, it does not add anything and we don't need to keep this updated:
```suggestion
Work in Progress, Internet-Draft.
```

IETF-OCM-MLS.md
```diff
@@ -0,0 +1,1914 @@
...
+# Discovery
+
+A server signals support for federation shares by including
+`"federation"` in the `shareTypes` array for a given resource type in
+its OCM discovery document at `/.well-known/ocm`:
+
+~~~ json
+{
+  "enabled": true,
+  "apiVersion": "1.4.0",
```
glpatcern, line 365: Are you sure? :-)

glpatcern (commented):
Just a couple of comments for now. I skimmed through the document and it's more or less fine, we can look into the details as part of the WG adoption process.

IETF-OCM-MLS.md
mickenordin: I did this because of authortools I-D nits complaining, but I agree it is a bit annoying, and will not cause a real issue.

IETF-OCM-MLS.md
mickenordin: It needs to say _something_ :sweat_smile:

glpatcern (approved):

PR #378, opened 2026-06-21
Prepare version 1.4

glpatcern:
Added missing cross references to the new I-Ds and prepared changelog

### DIFF
--- a/IETF-OCM.md
+++ b/IETF-OCM.md
@@ -839,7 +839,11 @@ MUST expose the `exchange-token` capability and a `tokenEndPoint`,
 because WebApp access requires the Receiving Server to exchange
 `protocol.webapp.sharedSecret` before presenting the WebApp to the
 browser.  If the Sending Server cannot offer this code flow, it MUST NOT
-include `protocol.webapp` in the notification.
+include `protocol.webapp` in the notification.  A Sending Server MAY
+serve Web apps either from the same hosting infrastructure or from
+external servers in the same organization: to facilitate the integration
+of external servers, the RECOMMENDED reference implementation is
+described at [OCM-IP].
 
 ## Fields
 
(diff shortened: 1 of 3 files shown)


IETF-OCM.md
```diff
@@ -839,7 +839,10 @@ MUST expose the `exchange-token` capability and a `tokenEndPoint`,
 because WebApp access requires the Receiving Server to exchange
 `protocol.webapp.sharedSecret` before presenting the WebApp to the
 browser.  If the Sending Server cannot offer this code flow, it MUST NOT
-include `protocol.webapp` in the notification.
+include `protocol.webapp` in the notification.  A Sending Server MAY
+serve Web apps either from itself or from external servers in the same
+organization: to facilitate the integration of external servers,
+a reference implementation is provided at [OCM-IP].
```
mickenordin: I think we should RECOMMEND that implementors who want interoperability use OCM-IP

IETF-OCM.md
```diff
@@ -885,8 +888,8 @@ include `protocol.webapp` in the notification.
   Server.
   In the federation case, OCM Servers MAY resolve the actual
   recipients by either querying external AAI systems, or exchanging
-  the groups' metadata between themselves.  Such exchange is out of
-  scope for this version of the this specification.
+  the groups' metadata between themselves.  For a reference
```
mickenordin: Same thing here, I think federated groups should be the RECOMMENDED way now

mickenordin (changes requested):

IETF-OCM.md
glpatcern: Agree for both!

mickenordin (approved):

MahdiBaghbani (approved):

PR #380, opened 2026-06-23
Add IANA registries

mickenordin:
This PR adds IANA registires for:

* notifications
* resourceTypes
* shareTypes
* protocols
* share payloads

It also registers OCM-MLS notification entries.

The idea with the registries is to try to find where the extension points of OCM should be. We want the protocol to be extensible AND interoperable. Where the spec is unclear or under-developed interoperability suffers. If we add extension points to the specification with out making it possible to explain how to use that extension point, only proprietary single vendor extensions are possible.

So for example writing in the specification text, that other kinds of notifications are allowed and that implementors MUST ignore notifications they don't understand, is adding an extension point, that is guaranteed not to be interoperable. The registries make sure that we get both extensibility AND interoperabilities at the same time.

### DIFF
--- a/IETF-OCM-MLS.md
+++ b/IETF-OCM-MLS.md
@@ -51,9 +51,9 @@ capabilities for resources shared with federated groups.
 
 Open Cloud Mesh [OCM] currently supports sharing resources with
 individual users across federated servers and with groups on a single
-server.  [OCM] also defines a `shareType` of `"federation"` but does
-not further specify its semantics.  This document gives `"federation"`
-concrete semantics: a federated group identified by an
+server.  This document defines a new `shareType`, `"federation"`, and
+registers it in the "OCM Share Types" registry defined by [OCM].  A
+`"federation"` share is addressed to a federated group identified by an
 OCM Address such as `research-group@receiver.example.org` whose
 membership spans multiple OCM servers, with group state managed through
 the MLS [RFC9420] epoch mechanism.
@@ -1848,6 +1848,36 @@ the "providerId" field (see {{mls-notification-types}}):
    | MLS_APPLICATION   | Group | active | This document |
    | MLS_REJOIN        | Group | active | This document |
    +===================+=======+========+===============+
+~~~
+
+The following entry is to be registered in the "OCM Share Types"
+registry defined in [OCM], within the "Open Cloud Mesh (OCM)
+Parameters" group.  This document is the registering specification for
+the "federation" share type:
+
+~~~
+   +============+===============+
+   | Share Type | Reference     |
+   +============+===============+
+   | federation | This document |
+   +============+===============+
+~~~
+
+The following entries are to be registered in the "OCM Share Payloads"
+registry defined in [OCM], within the "Open Cloud Mesh (OCM)
+Parameters" group.  They extend the existing "webdav", "webapp", and
+"ssh" protocols to the "federation" share type; the wire format of the
+share payload for these combinations is completely specified by this
+document together with [OCM] (see {{share-creation}}).  These
+registrations do not modify the protocols' own registrations:
+
+~~~
+   +===============+============+=====================+===============+
+   | Resource Type | Share Type | Protocols           | Reference     |
+   +===============+============+=====================+===============+
+   | file          | federation | webdav, webapp, ssh | This document |
+   | folder        | federation | webdav, webapp, ssh | This document |
+   +===============+============+=====================+===============+
 ~~~
 
 # Open Issues
--- a/IETF-OCM-MLS.md
+++ b/IETF-OCM-MLS.md
@@ -51,9 +51,9 @@ capabilities for resources shared with federated groups.
 
 Open Cloud Mesh [OCM] currently supports sharing resources with
 individual users across federated servers and with groups on a single
-server.  The specification also defines a `shareType` of `"federation"`
-but does not further specify its semantics.  This document gives
-`"federation"` a concrete definition: a federated group identified by an
+server.  [OCM] also defines a `shareType` of `"federation"` but does
+not further specify its semantics.  This document gives `"federation"`
+concrete semantics: a federated group identified by an
 OCM Address such as `research-group@receiver.example.org` whose
 membership spans multiple OCM servers, with group state managed through
 the MLS [RFC9420] epoch mechanism.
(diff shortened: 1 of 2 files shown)


IETF-OCM-MLS.md
```diff
@@ -1833,6 +1833,42 @@ defined in [RFC9420] Section 17.3:
...
+   +===============+============+==========+===============+
+   | file          | federation | webdav   | This document |
+   | file          | federation | webapp   | This document |
+   | file          | federation | ssh      | This document |
+   | folder        | federation | webdav   | This document |
+   | folder        | federation | webapp   | This document |
+   | folder        | federation | ssh      | This document |
+   +===============+============+==========+===============+
+~~~
+
```
glpatcern: This seems a duplicate of what is in `IETF-OCM.md`

IETF-OCM.md
```diff
@@ -1535,6 +1556,164 @@ Values" registry (using the template from [RFC9553]).
...
+
+IANA is requested to create the "OCM Share Payloads" registry in the
+"Open Cloud Mesh (OCM) Parameters" group.  Each entry records a
+combination of resource type, share type, and protocol for which a
+complete share payload is specified, together with a reference to the
+document that specifies that payload.
+
+The values in the "Resource Type", "Share Type", and "Protocol"
+columns MUST already appear in the "OCM Resource Types", "OCM Share
+Types", and "OCM Protocols" registries respectively.  For this
```
glpatcern: Given this, I don't really get why we'd need an extra dedicated registry for the Cartesian product of all other registries. If really needed, we'd have to also add all combinations where share type = `federation`.

IETF-OCM.md
```diff
@@ -1535,6 +1556,164 @@ Values" registry (using the template from [RFC9553]).
...
+~~~
+   +===========================+=======+==============+===============+
+   | Notification Type         | Scope | Status       | Reference     |
+   +===========================+=======+==============+===============+
+   | SHARE_ACCEPTED            | Share | active       | This document |
+   | SHARE_DECLINED            | Share | active       | This document |
+   | SHARE_UNSHARED            | Share | active       | This document |
+   | REQUEST_RESHARE           | Share | experimental | This document |
+   | RESHARE_UNDO              | Share | experimental | This document |
+   | RESHARE_CHANGE_PERMISSION | Share | experimental | This document |
```
glpatcern, line 1725: I think we also have `USER_REMOVED`, though I realize we mention it only in `spec.yaml`, not (yet) in the I-D. That's pending on the notifications work...

glpatcern (changes requested):
Thanks Micke, I have a couple of fixes and a conceptual question

IETF-OCM-MLS.md
mickenordin: This is registering the federation share type

IETF-OCM.md
mickenordin: Yes, what I want is to add the regestires, but I fully expect that the content change as we fix notifications.

IETF-OCM.md
mickenordin: I put user and group here, and federation in OCM-MLS, which is why it is missing here.

For the other part of the question, the problem I have is that the resourceType, shareType and protocol needs to be specified together, for the share payload to make sense. You cant just mix and match them. On the surface it looks like you can for shareType and resourceType, i.e. it makes sense to have any combination of file and folder along with a user, group or federation, but will it always really be possible to share any possible share with a group or federation? For ssh, we almost specified only one ssh key in the response, instead of an array, which would have limited the possibility to set any other resourceType other than user, and may still, in practice, limit the possibility to share with a federation. For other future protocols, this may be a legitimate limitation you would want to make.

In any case protocol is very much tied to reasourceType, in the sense that I would find it strange to share a calendar event over ssh, and I don't know what to do with that, since it is how the OCM-protocol is designed. A calendar event could well be shared as as vcard or jscal _file_ over ssh, sure, but not as a _calendar event_, because then caldav or jmap would make sense as protocols, but not ssh.

So I think it _looks_ strange now, since the base protocol is so file centric, but the use case is for other things, besides files and folders, to be registered.

With that being said, I am willing to be proven wrong here, because I don't like it either, but my feeling is that this is a limitation of the protocol, rather than a problem with the registries.

mickenordin: Oh, and I think there is one thing here, which I should make clear. The resourceType, shareType and protocol regisitries specifies what goes in the discovery, where as the share payload registry specifies what goes in the wire format of the share payload.

IETF-OCM.md
```diff
@@ -1647,42 +1647,56 @@ field of a [Share Creation Notification](#share-creation-notification).
...
+specifications MAY register additional combinations, including ones
+that extend an already-registered protocol to a new resource type or
+share type; doing so does not modify that protocol's own registration.
+The federation combinations are registered in this way by [OCM-MLS].
+If someone wants to specify how to share calendar events over ssh in an
+interoperable way, they can do so using this very mechanism.
+
+A Sending Server MUST NOT send a share for a combination that the
+Receiving Server does not advertise in Discovery (see
+[Share Creation Notification](#share-creation-notification)).
```
glpatcern: I think we should allow `webdav` as a legacy exception

IETF-OCM.md
```diff
@@ -1647,42 +1647,56 @@ field of a [Share Creation Notification](#share-creation-notification).
...
+The registered combinations are a constrained subset, not the full
+Cartesian product of those three registries, even though for the
+initial content the subset and the Cartesian product is the same thing.
+However, in other cases that file sharing, a protocol may only be
+meaningful for certain resource types.  A calendar event, for example,
+is usually shared over CalDAV or JMAP, not over ssh.  Other
+specifications MAY register additional combinations, including ones
+that extend an already-registered protocol to a new resource type or
+share type; doing so does not modify that protocol's own registration.
+The federation combinations are registered in this way by [OCM-MLS].
```
glpatcern, line 1681: I understand the concept, but I disagree with this. What about adding `federation` here, where `Reference` is populated with `This document and [OCM-MLS]`, and drop it from the other I-D?
My point is we do allow `federation` also as share payload, even if it's not specified, so the type belongs here, the internals belong to [OCM-MLS].

glpatcern (commented):
Some more comments. I start to get the principle, that is we define a way for implementers to extend the spec in the parts that may be extended, without needing to touch the core spec. This is very good in view of a final RFC.

glpatcern (dismissed):
I think the current state is the best solution. It seems we need some IANA expert to decide whether what is here makes sense.

mickenordin:
@glpatcern take a look now, I hope this is an acceptable interpretation of the consensus I think we reached over chat?

IETF-OCM.md
```diff
@@ -894,23 +894,19 @@ described in [OCM-IP].
...
-  a Directory Service, including at least one user at the Receiving
-  Server.
-  In the federation case, OCM Servers MAY resolve the actual
-  recipients by either querying external AAI systems, or exchanging
-  the groups' metadata between themselves.  For the latter, the
-  RECOMMENDED implementation is based on the MLS protocol and it is
-  described in [OCM-MLS].
-  Alternatively, the Receiving Server MAY hold the federated groups'
-  metadata and act as an OCM proxy, forwarding the OCM requests to
-  the actual members of the federation.
```
glpatcern, line 897: I note that we loose this comment. Not a blocker for now but it would be nice to bring it back, possibly in the OCM-MLS doc

glpatcern (dismissed):
Yes, this resolves my self-consistency concern. There's one comment that we may address later, and I have slightly rephrased the Discovery part to consistently refer to all registries. Please have a look and merge if you're happy with it.

mickenordin:
> Yes, this resolves my self-consistency concern. There's one comment that we may address later, and I have slightly rephrased the Discovery part to consistently refer to all registries. Please have a look and merge if you're happy with it.

You need to re-approve for merging, I removed a duplicate `calendar` that had snuck in.

glpatcern (approved):

PR #382, opened 2026-07-09
Sorted the Terms and removed duplicates

glpatcern:


mickenordin (approved):
This looks good to me, I double checked no entry was lost, so only duplicates were removed.

PR #383, opened 2026-07-09
OCM/OCM-MLS: Standardize on base64url

mickenordin:
We use base64 in several places, in particular the invitation strings needs to be URL safe. This commit standardizes on base64url encoding in all places where we use base64, including in diagrams and in the IETF-OCM-MLS draft.

The example invite string:

YTU1YTk2NmUtMTVjMS00Y2I5LWEzOWQtNGU0YzU0Mzk5YmFmQGNsb3VkLmV4YW1wbGUub3Jn

can be decoded using standard GNU coreutils like this:

echo -n 'YTU1YTk2NmUtMTVjMS00Y2I5LWEzOWQtNGU0YzU0Mzk5YmFmQGNsb3VkLmV4YW1wbGUub3Jn' | basenc -d --base64url

and yields: a55a966e-15c1-4cb9-a39d-4e4c54399baf@cloud.example.org


glpatcern (approved):
Perfect, this is very unambiguous

thibmeu (commented):
much clearer. i don't know about the state of implementations. it might be worth checking if this is breaking and update them accordingly

MahdiBaghbani (approved):
I like it!

MahdiBaghbani:
I have to remind us that this can become a bit misused as we have seen in the Reva based platforms other than the CERNBox in the past year's EOSC efforts.

while its fine for the invite to be base64, its definitely not fine for the actual OCM identifier of `opaque_id@fqdn` to be in base64 format. and even worse in `<base64>@fqdn` format.

glpatcern:
> much clearer. i don't know about the state of implementations. it might be worth checking if this is breaking and update them accordingly

That's a fair point, will have to clarify this. As a matter of fact, for CERNBox the base64 logic was added in the web frontend (as a plain `btoa()` JS call, not URL-safe), and ownCloud and OpenCloud are likely doing the same as they inherited the code. I don't know about Nextcloud.

Yet, this PR goes in the right direction and I will proceed with merging it.

PR #384, opened 2026-07-09
Make invite acceptance response and request for share use http-sig always

mickenordin:


IETF-OCM.md
```diff
@@ -760,6 +760,59 @@ contain the following information about its OCM API:
...
+
+## Signing Requirements
+
+A signed request MUST cover at least the following Signature-Input
+components:
+
+* "@method"             - HTTP method
+* "@target-uri"         - full request URI (scheme, authority,
+                          path, query)
+* "content-digest"      - [RFC9530] digest of the body
```
thibmeu, line 816: to confirm: is the goal here to prevent tampering in transit?

in any case, one thing that might be useful, you can present an example with one (or a selection) of algorithms from https://www.iana.org/assignments/http-digest-hash-alg/http-digest-hash-alg.xhtml. I suspect `sha256` is the most common

IETF-OCM.md
```diff
@@ -760,6 +760,59 @@ contain the following information about its OCM API:
...
+* "content-length"      - message size
+* "date"                - timestamp
```
thibmeu: question: why sign date if you have `created`?

IETF-OCM.md
```diff
@@ -760,6 +760,59 @@ contain the following information about its OCM API:
...
+
+The Signature-Input parameters MUST include `created`.
+
+A request signed in the context of OCM MUST include one and only one
+signature with the label `ocm` in its Signature and Signature-Input
+headers.
+
+A symmetric signing algorithm MUST NOT be used to sign the request, as
```
thibmeu: worth recommending a specific algorithm from [IANA registry](https://www.iana.org/assignments/http-message-signature/http-message-signature.xhtml#signature-algorithms) ? I believe that you want to exclude HMAC specifically

IETF-OCM.md
```diff
@@ -760,6 +760,59 @@ contain the following information about its OCM API:
...
+components:
+
+* "@method"             - HTTP method
+* "@target-uri"         - full request URI (scheme, authority,
+                          path, query)
+* "content-digest"      - [RFC9530] digest of the body
+* "content-length"      - message size
+* "date"                - timestamp
+
+The Signature-Input parameters MUST include `created`.
```
thibmeu: it's probably worth adding a `tag="ocm"` (or multiple tags depending on the steps, i'm not sure). per [section 2.3 of rfc9421](https://datatracker.ietf.org/doc/html/rfc9421#section-2.3-4.12)

> An application-specific tag for the signature as a String value. This value is used by applications to help identify signatures relevant for specific applications or protocols.

This would allow to filter by purpose (tag) instead of per label

IETF-OCM.md
```diff
@@ -760,6 +760,59 @@ contain the following information about its OCM API:
...
+
+A request signed in the context of OCM MUST include one and only one
+signature with the label `ocm` in its Signature and Signature-Input
+headers.
+
+A symmetric signing algorithm MUST NOT be used to sign the request, as
+the Receiving Server would not be able to verify the signature without
+having access to the shared secret in advance.
+
+## Verification Requirements
```
thibmeu, line 847: adding a short mention of `Accept-Signature` ([Section 5.1 of RFC 9421](https://datatracker.ietf.org/doc/html/rfc9421#section-5.1)) as a wayn for verifier to state which parameters they require and support seems relevant in this section

IETF-OCM.md
mickenordin: It has been a while since I wrote the http-sig part, butif I remember correctly, created is part of the signature it self and specifies when it was signed: https://www.rfc-editor.org/info/rfc9421/#signature-params

And date is the standard http header for when the request was made: https://www.rfc-editor.org/info/rfc9421/#section-7.2.2

So there may be a reason to make sure that these two fields do not skew too much to protect against replay attacks, and signing date is a good thing in that regard.

IETF-OCM.md
mickenordin: Since OCM is a server to server protocol, there is a single sending server and a single receiving server and my thinking is that nothing should need to interfere with the OCM request, a proxy does not need to add anything for the OCM request for instance, so having multiple signatures from different software regarding OCM only makes the http-sig implementation harder, that is why I chose to have one and only one label for OCM messages. I don't know if tag is a better mechanism than label? I think the reasoning would be the same, that we only want one signature. Is there a reason to use tag instead of, or in addition to label, and can you see a reason to have multiple of either for the purpose of OCM?

thibmeu: As I understand it, the quoted section of the RFC above specifically target the application use case.

One reason that you likely want to use a tag is that it's included under the signature base, while labels are not.
Labels may be rewritten by intermediaries (I agree they should not, but that's something which is allowed), therefore having a tag is better.

IETF-OCM.md
thibmeu: The consideration is mostly [7.2.4 of RFC 9421](https://www.rfc-editor.org/info/rfc9421/#section-7.2.4-2.2.1). The question is mostly that if that may allow for skew detection but it's not enforced, then probably not worth having "date" as a MUST, but rather have it as a SHOULD (or even left out to implementations)

Note: this is mostly an observation. You have a closer interaction with the implementation and that's really a clarifying question

mickenordin: I think you are right, but is it better to drop date than to enforce detection?

IETF-OCM.md
mickenordin: Ok, that sounds like a really good argument. So switching to one and only one tag instead of a lable would be the right move here?

mickenordin: Right, we need both the label and the tag I think

IETF-OCM.md
thibmeu: that'd be my preference, but that's something to sync with the group really

IETF-OCM.md
thibmeu: i think so. not sure if you want different tag for different ocm steps. tbd with the group

mickenordin: Does this make sense to you?

https://github.com/cs3org/OCM-API/pull/384/changes/6029e51b48183f374a491b1221f27f3426bd6087

IETF-OCM.md
mickenordin: This should be fixed now

IETF-OCM.md
mickenordin: Is this needed since we have the http-sig capability in the discovery, already negotiating this? Do you know if it is usually is used by software libraries to prepare signatures, or if it would be redundant in this context given our discovery mechanism?

IETF-OCM.md
glpatcern: It does to me, good to know about this subtle difference between tags and labels

IETF-OCM.md
mickenordin: Yes, good I recommend ed25519 now

IETF-OCM.md
mickenordin: Yes, good.

mickenordin:
I have now tried to describe the consensus about signature applicability and backwards compatibility in such a way that the spec allows incremental adoption.

glpatcern (approved):
Modulo the last unresolved conversation, for which I don't have a strong opinion, this change is good to go and very welcome for me.

PR #390, opened 2026-07-22
Rework OCM-API diagrams, sync OpenAPI to draft

MahdiBaghbani:
This makes the OCM-API spec easier to navigate and to implement from.
Three things landed here, all in service of the same goal: help implementers and users find the normative home for each concept.

- A navigation index and a signing direction index in the I-D, so the four
  orthogonal planes and the per-endpoint signing rules have a single pointer
  table.
- A reworked `diagrams/` folder: 8 numbered diagrams plus a `README.md`,
  with a unified palette and a style contract.
- `spec.yaml` aligned to the current I-D draft.

The part that made me stop and think for some time is that the spec described the four planes (discovery, share, peer interaction, enforcement) only implicitly, and the diagrams did not share a vocabulary or palette.

An implementer reading `IETF-OCM.md` had no single place to land.
I think this is the right direction: keep the draft normative, add informative pointers and diagrams, and make the OpenAPI match the draft.

No new normative rules are introduced by the diagrams or the indexes


IETF-OCM.md
```diff
@@ -2546,6 +2611,11 @@ process and it shall be removed when going to RFC last call.
 The complete changelog is updated in the OCM-API GitHub repository.
 
 ## Version 06
+* Added informative aids: same-string note for `must-exchange-token`,
```
glpatcern, line 2612: This now should go to a `## Version 07` section

spec.yaml
```diff
@@ -379,20 +379,20 @@ components:
...
+                  registry (see the IANA Considerations section of the OCM
+                  Internet-Draft).
                 example: file
               shareTypes:
                 type: array
                 description: |
                   The supported recipient share types.
-                  MUST contain `"user"` at a minimum, plus optionally
-                  `"group"` and `"federation"`.
+                  MUST contain `user` at a minimum, plus optionally `group` or
```
glpatcern: As discussed over chat, here we need to revert to the previous text. The understanding is that `spec.yaml` maps to the _union_ of the Internet-Drafts, not to the core only.
I'm undecided whether we want to quote here the IANA registries: eventually, I think we want a very minimal description that points to the I-Ds rather than duplicating too much text.

spec.yaml
```diff
@@ -577,9 +574,12 @@ components:
         shareWith:
           type: string
           description: >
-            OCM Address of the user, group or federation the
-            provider wants to share the resource with. This MUST be known
-            in advance, either via a previous Invitation or through other means.
+            OCM Address of the user or group the provider wants to share the
```
glpatcern: Same here, let's keep the original text and example

glpatcern (changes requested):
This is all very good, in particular the diagrams are excellent!
As discussed, some relatively minor changes on the yaml

glpatcern (approved):
Looks good! In particular the diagrams are now much more descriptive, I think some adjustments can still be done but it's a very good progress already.

PR #391, opened 2026-07-23
Address reviewer feedback on httpsig

mickenordin:
## Address HTTPSig review feedback

This PR addresses the review feedback from Justin Richer on the
HTTP Message Signatures section of the draft.

### Changes

**Date header** 

The `Date` header is no longer covered by signatures: intermediaries commonly rewrite it, and the `created` signature parameter already conveys creation time (RFC 9421 §7.2.4).

**Algorithm derivation**

The signature algorithm is now derived from the key material instead of being restricted to the IANA "HTTP Signature Algorithms" registry. Each JWK MUST carry `kid` and `alg`; the JWK `alg` MUST be an asymmetric JOSE-registered algorithm and SHOULD be fully-specified per RFC 9864 (`Ed25519` RECOMMENDED, which also replaces the RFC 8037 reference). The algorithm is applied per RFC 9421 §3.3.7, so new (e.g. post-quantum) JOSE algorithms work without updating this document. `none` and MACs are forbidden. The HTTPSig `alg` signature parameter is OPTIONAL; if present it MUST denote the same algorithm as the JWK, and is omitted when no registry  counterpart exists. A new "Keys and Algorithms" subsection holds this text.

**JWKS discovery**

The unregistered `/.well-known/jwks.json` endpoint is replaced by a `jwksUri` field in the Discovery response (also added to `spec.yaml`). Servers advertising the `http-sig` capability MUST provide it.

**keyid**

The `keyid` signature parameter is now REQUIRED and MUST equal the `kid` of the verification key in the signer's JWK Set; verifiers reject signatures with no matching key.

**Informative examples**

Appendix B is now explicitly informative; its normative statements were moved into the HTTP Message Signatures section. All signature examples use RFC 8792 line wrapping, and the
Token Request example uses `Content-Digest` (RFC 9530) instead of the legacy `Digest` header, with real digest values.

**Consistency pass** 

The Trusted Server definition, Discovery introduction, and Decision to Discard now speak of JWK Sets discovered via `jwksUri` and the `keyid`/`kid` match; the Token Request example covers `content-length` as required; the Appendix B example carries a real `Content-Digest` and `Content-Length`.

The Version 07 changelog records all of the above.

### DIFF
--- a/IETF-OCM.md
+++ b/IETF-OCM.md
@@ -771,12 +771,12 @@ contain the following information about its OCM API:
   `"/index.php/apps/sciencemesh/accept"` is specified here then a WAYF
   Page SHOULD redirect the end-user to `/index.php/apps/sciencemesh/
   accept?token=zi5kooKu3ivohr9a&providerDomain=cloud.example.org`.
-* OPTIONAL: jwksUri (string) - URL of a JWK Set document [RFC7517]
-  containing the public keys this OCM Server uses for HTTP Message
-  Signatures.  The URL is discovered from this field; it is not a
-  fixed path in the OCM API.
+* OPTIONAL: jwksUri (string) - https URL of a JWK Set document
+  [RFC7517] containing the public keys this OCM Server uses for HTTP
+  Message Signatures.  The URL is discovered from this field; it is
+  not a fixed path in the OCM API.
   Implementations that advertise the `"http-sig"` capability MUST
-  provide this URL as well.
+  provide this URL as well, and it MUST use https.
   Example: `"https://cloud.example.org/ocm/jwks"`.
 * OPTIONAL: tokenEndPoint (string) - URL of the token endpoint hosted by
   this OCM Server.  When this OCM Server acts as Sending Server, the
@@ -837,16 +837,17 @@ components:
 * "content-digest"      - [RFC9530] digest of the body
 * "content-length"      - message size
 
-The Signature-Input parameters MUST include `created` and `keyid`.
+The signature parameters MUST include `created` and `keyid`.
 Freshness and replay protection are anchored on `created` (see
 Verification Requirements).  The `keyid` value MUST be equal to the
 `kid` value of the corresponding key in the signer's JWK Set (see
 [Keys and Algorithms](#keys-and-algorithms)).
 
 The `Date` header is deliberately not covered by the signature:
-intermediaries commonly rewrite it, which would make signatures
-fragile, and the `created` signature parameter already conveys the
-message's creation time (see Section 7.2.4 of [RFC9421]).
+intermediaries sometimes rewrite it (see Section 6.6.1 of [RFC9110]),
+which would make signatures fragile, and the `created` signature
+parameter already conveys the message's creation time (see
+Section 7.2.4 of [RFC9421]).
 
 The `content-digest` component binds the request body to the signature,
 protecting it against modification in transit.  Its value MUST use a
@@ -1280,15 +1281,18 @@ A 503 response status means that the Receiver is temporary unavailable.
 The Receiving Server MAY discard the notification if any of the
 following hold true:
 
-* the HTTP Signature is missing but the Sending Server advertises the
-  `http-sig` capability in the Discovery response obtained from the
-  FQDN part of the `sender` field in the request body
-* the HTTP Signature is missing
-* the HTTP Signature is not valid (see [HTTP Message
+* the HTTP Message Signature is missing but the Sending Server
+  advertises the `http-sig` capability in the Discovery response
+  obtained from the FQDN part of the `sender` field in the request
+  body
+* the Sending Server advertises the `http-sig` capability but its
+  Discovery response carries no `jwksUri`
+* the HTTP Message Signature is missing
+* the HTTP Message Signature is not valid (see [HTTP Message
   Signatures](#http-message-signatures))
 * no trusted JWK Set can be obtained for the FQDN part of the
   `sender` field in the request body, via the `jwksUri` field of its
-  Discovery response or otherwise
+  Discovery response
 * that JWK Set contains no key whose `kid` equals the `keyid`
   signature parameter (see [HTTP Message
   Signatures](#http-message-signatures))
@@ -2064,6 +2068,10 @@ Key Words](https://datatracker.ietf.org/html/rfc8174)", May 2017.
 [RFC8615] Nottingham, M. "[Well-Known Uniform Resource Identifiers
 (URIs)](https://datatracker.ietf.org/doc/html/rfc8615)", May 2019
 
+[RFC9110] Fielding, R., Nottingham, M. and Reschke, J. "[HTTP
+Semantics](https://datatracker.ietf.org/doc/html/rfc9110)",
+June 2022.
+
 [RFC9421] Backman, A., Richer, J. and Sporny, M. "[HTTP Message
 Signatures](https://tools.ietf.org/html/rfc9421)", February 2024.
 
@@ -2075,7 +2083,8 @@ Representation of Contact Data](
 https://datatracker.ietf.org/doc/html/rfc9553), May 2024"
 
 [RFC9864] Jones, M., Steele, O., "[Fully-Specified Algorithms for
-JOSE and COSE](https://datatracker.ietf.org/doc/html/rfc9864)",
+JSON Object Signing and Encryption (JOSE) and CBOR Object Signing
+and Encryption (COSE)](https://datatracker.ietf.org/doc/html/rfc9864)",
 October 2025.
 
 ## Informative References
@@ -2122,7 +2131,7 @@ out of scope for this specification: a mechanism similar to the
 
 This appendix is informative.
 
-## JWKS Endpoint
+## Published JWK Set
 
 An OCM Server that advertises the `http-sig` capability publishes
 its public keys, in the format specified by [RFC7517], at the URL
@@ -2138,7 +2147,7 @@ an example response from `https://sender.example.org/ocm/jwks`:
       "crv": "Ed25519",
       "alg": "Ed25519",
       "kid": "sender.example.org#key1",
-      "x": "11qYAYKxCrfVS_7TyWQHOg7hcvPapiMlrwIaaPcHURo"
+      "x": "AzeStPAdQ-vbi6bJVxcQjBPF2I7fDZdfDZ_XZ3J4Azs"
     }
   ]
 }
@@ -2184,10 +2193,10 @@ NOTE: '\' line wrapping per RFC 8792
 
 "@method": POST
 "@target-uri": https://receiver.example.org/ocm/shares
-"content-digest": sha-256=:[digest-value]:
-"content-length": [body-length]
+"content-digest": sha-256=:/Uz47cfD1HOdMTIqcWjd84iMLQ4gVJdC7ZFACf1ViqU=:
+"content-length": 521
 "@signature-params": ("@method" "@target-uri" "content-digest" \
-  "content-length");created=[timestamp];\
+  "content-length");created=1785060500;\
   keyid="sender.example.org#key1";alg="ed25519";tag="ocm"
 </sourcecode>
 
@@ -2200,12 +2209,13 @@ base above:
 <sourcecode type="http">
 NOTE: '\' line wrapping per RFC 8792
 
-Content-Digest: sha-256=:[digest-value]:
-Content-Length: [body-length]
+Content-Digest: sha-256=:/Uz47cfD1HOdMTIqcWjd84iMLQ4gVJdC7ZFACf1ViqU=:
+Content-Length: 521
 Signature-Input: sig1=("@method" "@target-uri" "content-digest" \
-  "content-length");created=[timestamp];\
+  "content-length");created=1785060500;\
   keyid="sender.example.org#key1";alg="ed25519";tag="ocm"
-Signature: sig1=:[signature-value]=:
+Signature: sig1=:Epax4XFqvbrrDzQUAaNl+uECasEKbArgz8AlEkXspvYX2fSkQFlC\
+  kaXImZnYU898C2t0IdmbFhjR8bkN0hZgCg==:
 </sourcecode>
 
 See [HTTP Message Signatures](#http-message-signatures) for the
@@ -2221,7 +2231,9 @@ illustrates the procedure to verify an incoming signed request:
    request body
 2. Fetch the Discovery response from
    `https://<provider-domain>/.well-known/ocm` and read its
-   `jwksUri` field
+   `jwksUri` field.  If the Sending Server advertises the `http-sig`
+   capability but no `jwksUri` is present, the receiver MUST reject
+   the request as non-conformant
 3. Fetch the public keys from the URL given by `jwksUri`
 4. Locate the unique signature carrying the `tag="ocm"` parameter in
    the `Signature-Input` header, disregarding its dictionary label
@@ -2238,7 +2250,7 @@ illustrates the procedure to verify an incoming signed request:
 ## Validating the Payload
 
 Following the validation of the signature, the host also confirms
-the validity of the payload, as required under Verification
+the validity of the payload, as specified under Verification
 Requirements in
 [HTTP Message Signatures](#http-message-signatures).
 
--- a/IETF-OCM.md
+++ b/IETF-OCM.md
@@ -1504,8 +1504,8 @@ for display purposes only):
 POST {tokenEndPoint} HTTP/1.1
 Host: cloud.example.org
 Content-Type: application/x-www-form-urlencoded
-Digest: SHA-256=ok6mQ3WZzKc8nb7s/Jt2yY1uK7d2n8Zq7dhl3Q0s1xk=
-Content-Length: 101
+Content-Digest: sha-256=:81kCnlO5UY/mZ8UgpxBWnq18GY3WhzJnDjOTvSvjbhw=:
+Content-Length: 80
 Signature-Input:
   sig1=("@method" "@target-uri" "content-digest");
   created=1730815200;
--- a/IETF-OCM.md
+++ b/IETF-OCM.md
@@ -834,9 +834,11 @@ components:
 * "content-digest"      - [RFC9530] digest of the body
 * "content-length"      - message size
 
-The Signature-Input parameters MUST include `created`.  Freshness and
-replay protection are anchored on `created` (see Verification
-Requirements).
+The Signature-Input parameters MUST include `created` and `keyid`.
+Freshness and replay protection are anchored on `created` (see
+Verification Requirements).  The `keyid` value MUST be equal to the
+`kid` value of the corresponding key in the signer's JWK Set (see
+[Keys and Algorithms](#keys-and-algorithms)).
 
 The `Date` header is deliberately not covered by the signature:
 intermediaries commonly rewrite it, which would make signatures
@@ -905,8 +907,9 @@ registry, the `alg` signature parameter MUST be omitted.
 ## Verification Requirements
 
 Verifiers MUST reject signatures that omit any of the mandatory
-components listed under Signing Requirements or the `created`
-parameter, and MUST reject signatures whose `created` value is more
+components listed under Signing Requirements or the `created` or
+`keyid` parameters, and MUST reject signatures whose `created` value
+is more
 than a small implementation-defined skew tolerance in the future, or
 older than the verifier's freshness window.
 
@@ -914,8 +917,9 @@ A `Content-Digest` header value carrying multiple algorithms MUST have
 every recognised digest match the body; a single match alongside a
 recognised mismatch MUST be treated as an integrity failure.
 
-Verifiers MUST reject a signature if the JWK identified by `keyid`
-does not carry an acceptable `alg` value (see
+Verifiers MUST reject a signature if the signer's JWK Set contains
+no key whose `kid` equals the `keyid` parameter, if the JWK
+identified by `keyid` does not carry an acceptable `alg` value (see
 [Keys and Algorithms](#keys-and-algorithms)), or if an `alg`
 signature parameter is present and does not correspond to the
 algorithm derived from that JWK.
@@ -2668,6 +2672,8 @@ process and it shall be removed when going to RFC last call.
 The complete changelog is updated in the OCM-API GitHub repository.
 
 ## Version 07
+* Required the `keyid` signature parameter and that it matches the
+  `kid` of the verification key in the signer's JWK Set.
 * The HTTP Message Signature algorithm is now derived from the JWK
   identified by `keyid`, per Section 3.3.7 of [RFC9421], instead of
   being restricted to the "HTTP Signature Algorithms" registry; the
--- a/IETF-OCM.md
+++ b/IETF-OCM.md
@@ -791,10 +791,11 @@ described in the respective sections.  This section specifies the
 normative requirements for producing and verifying those signatures.
 Appendix B contains a complete example.
 
-Public keys for signature verification are published as a JWK Set
-[RFC7517] at the URL advertised in the `jwksUri` field of the
-signer's [Discovery](#ocm-api-discovery) response, when the
-`http-sig` capability is advertised.
+Public keys for signature verification are distributed as follows:
+an OCM Server that advertises the `http-sig` capability MUST
+publish its public keys as a JWK Set [RFC7517] at the URL
+advertised in the `jwksUri` field of its
+[Discovery](#ocm-api-discovery) response.
 
 ## Applicability
 
@@ -935,6 +936,11 @@ discretion, or rejected when the receiver advertises
 `must-use-http-sig`).  Signatures without `tag="ocm"` MAY coexist (e.g.
 proxy-attached signatures) but verifiers MUST NOT process them as part
 of OCM signature processing.
+
+After successful signature verification, the verifier SHOULD confirm
+that the payload is consistent with the signer, i.e. that the
+actions implied by the payload were initiated on behalf of the
+origin of the request.
 
 ## Signing Direction Index
 
@@ -2109,11 +2115,14 @@ out of scope for this specification: a mechanism similar to the
 
 # Appendix B: JWKS and HTTP Signature Examples
 
+This appendix is informative.
+
 ## JWKS Endpoint
 
-An OCM Server that advertises the `http-sig` capability MUST publish
+An OCM Server that advertises the `http-sig` capability publishes
 its public keys, in the format specified by [RFC7517], at the URL
-advertised in the `jwksUri` field of its Discovery response.  Here is
+advertised in the `jwksUri` field of its Discovery response (see
+[HTTP Message Signatures](#http-message-signatures)).  Here is
 an example response from `https://sender.example.org/ocm/jwks`:
 
 ~~~
@@ -2196,10 +2205,8 @@ Signature-Input: sig1=("@method" "@target-uri" "content-digest"
 Signature: sig1=:[signature-value]=:
 </sourcecode>
 
-The covered components, the `created` parameter, the single `ocm`
-tag, and the prohibition on symmetric algorithms shown here are
-normative; see [HTTP Message Signatures](#http-message-signatures) for
-the full requirements.
+See [HTTP Message Signatures](#http-message-signatures) for the
+normative requirements illustrated by this example.
 
 ## Verifying a Signature (Receiver)
 
@@ -2225,10 +2232,10 @@ illustrates the procedure to verify an incoming signed request:
 
 ## Validating the Payload
 
-Following the validation of the signature, the host SHOULD also confirm
-the validity of the payload, that is ensuring that the actions implied
-in the payload actually initiated on behalf of the source of the
-request.
+Following the validation of the signature, the host also confirms
+the validity of the payload, as required under Verification
+Requirements in
+[HTTP Message Signatures](#http-message-signatures).
 
 As an example, if the payload is about initiating a new share, the file
 owner has to be an account from the instance at the origin of the
--- a/IETF-OCM.md
+++ b/IETF-OCM.md
@@ -1497,27 +1497,24 @@ To obtain an access token, the Receiving Server MUST send an HTTP POST
 request to the Sending Server’s {tokenEndPoint} as discovered in the
 OCM provider metadata, following section 4.4.2 of [RFC6749].  The
 request payload MUST be in `x-www-form-urlencoded` form, as shown
-in the following example (with line breaks in the Signature headers
-for display purposes only):
+in the following example:
 
 <sourcecode type="http">
+NOTE: '\' line wrapping per RFC 8792
+
 POST {tokenEndPoint} HTTP/1.1
 Host: cloud.example.org
 Content-Type: application/x-www-form-urlencoded
 Content-Digest: sha-256=:81kCnlO5UY/mZ8UgpxBWnq18GY3WhzJnDjOTvSvjbhw=:
 Content-Length: 80
-Signature-Input:
-  sig1=("@method" "@target-uri" "content-digest");
-  created=1730815200;
-  keyid="receiver.example.org#key1";
-  alg="ed25519";
-  tag="ocm"
-Signature: sig1=:bM2sV2a4oM8pWc4Q8r9Zb8bQ7a2vH1kR9xT0yJ3uE4wO5lV6bZ1cP
-  2rN3qD4tR5hC=:
+Signature-Input: sig1=("@method" "@target-uri" "content-digest");\
+  created=1730815200;keyid="receiver.example.org#key1";\
+  alg="ed25519";tag="ocm"
+Signature: sig1=:bM2sV2a4oM8pWc4Q8r9Zb8bQ7a2vH1kR9xT0yJ3uE4wO5lV6bZ\
+  1cP2rN3qD4tR5hC=:
 
-grant_type=authorization_code&
-client_id=receiver.example.org&
-code=my_secret_code
+grant_type=authorization_code&client_id=receiver.example.org&code=\
+  my_secret_code
 </sourcecode>
 
 The request MUST be signed using an HTTP Message Signature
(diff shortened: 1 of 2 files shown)


spec.yaml
```diff
@@ -531,6 +531,17 @@ components:
...
+          type: string
+          format: uri
+          description: >
+            Optional URL of a JWK Set document [RFC7517] containing the
+            public keys this OCM Server uses for HTTP Message Signatures
+            [RFC9421]. The URL is discovered from this field; it is not a
+            fixed path in the OCM API. If the `http-sig` capability is
+            exposed, the jwksUri MUST be advertised in the discovery
+            response.
+          example: https://cloud.example.org/ocm/jwks
```
MahdiBaghbani, line 545: Could we require https for `jwksUri` here?

Right now it is `format: uri`, so `http://` is valid. Since this URL serves the verification keys, I think it should be https only, like `tokenEndPoint` arguably should be too.

```suggestion
        jwksUri:
          type: string
          format: uri
          pattern: "^https://"
          description: >
            Optional URL of a JWK Set document [RFC7517] containing the
            public keys this OCM Server uses for HTTP Message Signatures
            [RFC9421]. The URL is discovered from this field; it is not a
            fixed path in the OCM API. If the `http-sig` capability is
            exposed, the jwksUri MUST be advertised in the discovery
            response and MUST use https.
          example: https://cloud.example.org/ocm/jwks
```

IETF-OCM.md
```diff
@@ -769,6 +771,13 @@ contain the following information about its OCM API:
   `"/index.php/apps/sciencemesh/accept"` is specified here then a WAYF
   Page SHOULD redirect the end-user to `/index.php/apps/sciencemesh/
   accept?token=zi5kooKu3ivohr9a&providerDomain=cloud.example.org`.
+* OPTIONAL: jwksUri (string) - URL of a JWK Set document [RFC7517]
+  containing the public keys this OCM Server uses for HTTP Message
+  Signatures.  The URL is discovered from this field; it is not a
+  fixed path in the OCM API.
+  Implementations that advertise the `"http-sig"` capability MUST
+  provide this URL as well.
+  Example: `"https://cloud.example.org/ocm/jwks"`.
```
MahdiBaghbani: Same https point in the I-D text, so the two stay in sync.

```suggestion
* OPTIONAL: jwksUri (string) - https URL of a JWK Set document
  [RFC7517] containing the public keys this OCM Server uses for HTTP
  Message Signatures.  The URL is discovered from this field; it is
  not a fixed path in the OCM API.
  Implementations that advertise the `"http-sig"` capability MUST
  provide this URL as well, and it MUST use https.
  Example: `"https://cloud.example.org/ocm/jwks"`.
```

IETF-OCM.md
```diff
@@ -784,9 +793,10 @@ described in the respective sections.  This section specifies the
...
 Appendix B contains a complete example.
 
-Public keys for signature verification are published in the format
-specified by [RFC7517] at the signer's `/.well-known/jwks.json`
-endpoint, if the `http-sig` capability is included in the
+Public keys for signature verification are distributed as follows:
+an OCM Server that advertises the `http-sig` capability MUST
+publish its public keys as a JWK Set [RFC7517] at the URL
+advertised in the `jwksUri` field of its
 [Discovery](#ocm-api-discovery) response.
```
MahdiBaghbani, line 802: Removing the fixed path is a real change for peers that hardcoded it, and it adds a mandatory Discovery round-trip. I think a short compatibility note would help implementers transition.

```suggestion
advertised in the `jwksUri` field of its
[Discovery](#ocm-api-discovery) response.

Prior versions of this document published the JWK Set at the fixed
path `/.well-known/jwks.json`.  That path is no longer specified;
peers MUST obtain the JWK Set location from the `jwksUri` field of the
Discovery response.  Implementations that previously hardcoded the
fixed path MUST fetch Discovery and read `jwksUri` before verifying
a signature.
```

IETF-OCM.md
```diff
@@ -849,24 +863,70 @@ A request MUST include one and only one signature carrying
...
+see [Keys and Algorithms](#keys-and-algorithms).
+
+## Keys and Algorithms
+
+Each key published in the signer's JWK Set MUST include the `kid`
+and `alg` parameters [RFC7517].  The JWK `alg` value MUST identify
+an asymmetric signature algorithm registered in the IANA "JSON Web
+Signature and Encryption Algorithms" registry [RFC7518], and SHOULD
+be a fully-specified algorithm [RFC9864]; `Ed25519` ([RFC8032],
+[RFC9864]) is RECOMMENDED.
```
MahdiBaghbani, line 881: Could we add a `kid` uniqueness MUST here? Right now two keys in the same JWK Set can share a `kid`, and a verifier has no defined way to pick between them.

```suggestion
Each key published in the signer's JWK Set MUST include the `kid`
and `alg` parameters [RFC7517].  The `kid` of each key in a JWK Set
MUST be unique within that set.  The JWK `alg` value MUST identify
an asymmetric signature algorithm registered in the IANA "JSON Web
Signature and Encryption Algorithms" registry [RFC7518], and SHOULD
be a fully-specified algorithm [RFC9864]; `Ed25519` ([RFC8032],
[RFC9864]) is RECOMMENDED.
```

IETF-OCM.md
```diff
@@ -849,24 +863,70 @@ A request MUST include one and only one signature carrying
...
+determined by the JWK, as described above.  Note that this
+parameter takes its values from a different registry than the JWK
+`alg` value.  If the parameter is present, the algorithm it names
+in the IANA "HTTP Signature Algorithms" registry [IANA-SIG-ALG]
+MUST denote the same algorithm as the JWK `alg` value, for
+example, `ed25519` for a JWK with `alg` `Ed25519`, and verifiers
+MUST reject the signature otherwise
+(see Section 7 of [RFC9421] on algorithm confusion and substitution
+attacks).  When the JWK algorithm has no counterpart in that
+registry, the `alg` signature parameter MUST be omitted.
```
MahdiBaghbani, line 913: I think the cross-registry `alg` mapping needs to be spelled out for more than Ed25519, otherwise "MUST denote the same algorithm" is hard to apply for RSA-PSS or ECDSA.

My preference would be a small table. I left the registry strings below as I read them from the IANA HTTP Signature Algorithms page, but could you please double-check them before this lands.

```suggestion
The `alg` signature parameter is OPTIONAL: the algorithm is always
determined by the JWK, as described above.  Note that this
parameter takes its values from a different registry than the JWK
`alg` value.  If the parameter is present, the algorithm it names
in the IANA "HTTP Signature Algorithms" registry [IANA-SIG-ALG]
MUST denote the same algorithm as the JWK `alg` value, and verifiers
MUST reject the signature otherwise (see Section 7 of [RFC9421] on
algorithm confusion and substitution attacks).  When the JWK
algorithm has no counterpart in that registry, the `alg` signature
parameter MUST be omitted, and verifiers MUST reject the signature
if it is present.

The following table maps the JWK `alg` values this document allows
to their HTTP Signature Algorithm registry counterparts, where one
exists:

  JWK `alg`  | HTTP Signature `alg`
  ---------- | --------------------
  Ed25519    | ed25519
  Ed448      | ed448
  ES256      | ecdsa-p256-sha256
  ES384      | ecdsa-p384-sha384
  ES512      | ecdsa-p521-sha512
  PS256      | rsa-pss-sha256
  PS384      | rsa-pss-sha384
  PS512      | rsa-pss-sha512

For JWK `alg` values not listed above, the `alg` signature parameter
MUST be omitted, because the HTTP Signature Algorithms registry does
not name a counterpart.
```

IETF-OCM.md
```diff
@@ -849,24 +863,70 @@ A request MUST include one and only one signature carrying
...
 recognised mismatch MUST be treated as an integrity failure.
 
+Verifiers MUST reject a signature if the signer's JWK Set contains
+no key whose `kid` equals the `keyid` parameter, if the JWK
+identified by `keyid` does not carry an acceptable `alg` value (see
+[Keys and Algorithms](#keys-and-algorithms)), or if an `alg`
+signature parameter is present and does not correspond to the
+algorithm derived from that JWK.
+
 Verifiers MUST identify the OCM signature by its `tag="ocm"`
```
MahdiBaghbani, line 935: And the matching reject rule in Verification Requirements, so an ambiguous `kid` is a failure rather than undefined.

```suggestion
Verifiers MUST reject a signature if the signer's JWK Set contains
no key whose `kid` equals the `keyid` parameter, if more than one key
in the JWK Set carries that `kid`, if the JWK identified by `keyid`
does not carry an acceptable `alg` value (see
[Keys and Algorithms](#keys-and-algorithms)), or if an `alg`
signature parameter is present and does not correspond to the
algorithm derived from that JWK.
```

IETF-OCM.md
```diff
@@ -1215,16 +1280,18 @@ A 503 response status means that the Receiver is temporary unavailable.
...
+  FQDN part of the `sender` field in the request body
 * the HTTP Signature is missing
-* the HTTP Signature is not valid
-* no keypair is trusted or discoverable from the FQDN part of the
-  `sender` field in the request body
-* the keypair used to generate the HTTP Signature doesn't match the one
-  trusted or discoverable from the FQDN part of the `sender` field
-  in the request body
+* the HTTP Signature is not valid (see [HTTP Message
+  Signatures](#http-message-signatures))
```
MahdiBaghbani: Two things here: I think these should say "HTTP Message Signature" to match the section title and RFC 9421, and I think the discard list should call out "advertises `http-sig` but no `jwksUri`" as an explicit case so peers do not fall through to a MAY-discard.

```suggestion
* the HTTP Message Signature is missing but the Sending Server advertises the
  `http-sig` capability in the Discovery response obtained from the
  FQDN part of the `sender` field in the request body
* the Sending Server advertises the `http-sig` capability but its
  Discovery response carries no `jwksUri`
* the HTTP Message Signature is missing
* the HTTP Message Signature is not valid (see [HTTP Message
  Signatures](#http-message-signatures))
```

IETF-OCM.md
```diff
@@ -1215,16 +1280,18 @@ A 503 response status means that the Receiver is temporary unavailable.
...
+* no trusted JWK Set can be obtained for the FQDN part of the
+  `sender` field in the request body, via the `jwksUri` field of its
+  Discovery response or otherwise
```
MahdiBaghbani: The "or otherwise" channel is not defined anywhere, so I think it weakens the discard rule. My preference would be to drop it and keep the JWK Set reachable only via `jwksUri`.

```suggestion
  `sender` field in the request body, via the `jwksUri` field of its
  Discovery response
```

IETF-OCM.md
```diff
@@ -2148,24 +2219,28 @@ illustrates the procedure to verify an incoming signed request:
 
 1. Extract the provider domain from the `sender` field in the
    request body
-2. Fetch the public key from
-   `https://<provider-domain>/.well-known/jwks.json`
-3. Locate the unique signature carrying the `tag="ocm"` parameter in
+2. Fetch the Discovery response from
+   `https://<provider-domain>/.well-known/ocm` and read its
+   `jwksUri` field
+3. Fetch the public keys from the URL given by `jwksUri`
```
MahdiBaghbani, line 2244: Step 3 fetches keys from `jwksUri`, but it does not say what the receiver does when `http-sig` is advertised and `jwksUri` is missing. I think an explicit reject step would help.

```suggestion
1. Extract the provider domain from the `sender` field in the
   request body
2. Fetch the Discovery response from
   `https://<provider-domain>/.well-known/ocm` and read its
   `jwksUri` field.  If the Sending Server advertises the `http-sig`
   capability but no `jwksUri` is present, the receiver MUST reject
   the request as non-conformant
3. Fetch the public keys from the URL given by `jwksUri`
```

IETF-OCM.md
```diff
@@ -2148,24 +2219,28 @@ illustrates the procedure to verify an incoming signed request:
...
 ## Validating the Payload
 
-Following the validation of the signature, the host SHOULD also confirm
-the validity of the payload, that is ensuring that the actions implied
-in the payload actually initiated on behalf of the source of the
-request.
+Following the validation of the signature, the host also confirms
+the validity of the payload, as required under Verification
+Requirements in
+[HTTP Message Signatures](#http-message-signatures).
```
MahdiBaghbani, line 2262: I think "as required under Verification Requirements" reads as MUST, but the normative text is a SHOULD. Could we make the strength match?

```suggestion
Following the validation of the signature, the host also confirms
the validity of the payload, as specified under Verification
Requirements in
[HTTP Message Signatures](#http-message-signatures).
```

IETF-OCM.md
```diff
@@ -2609,6 +2684,21 @@ process and it shall be removed when going to RFC last call.
...
+  `alg` signature parameter is optional and checked for consistency.
+* Replaced the unregistered `/.well-known/jwks.json` endpoint with a
+  `jwksUri` field in the Discovery response, from which the location
+  of the JWK Set for HTTP Message Signatures is discovered.
+* The `Date` header is no longer covered by signatures: freshness is
+  anchored on the `created` signature parameter (see Section 7.2.4
+  of [RFC9421]).
+* Updated the signature examples: replaced the legacy `Digest` header
+  with `Content-Digest` [RFC9530] and applied [RFC8792] line
+  wrapping.
```
MahdiBaghbani, line 2730: I think the Version 07 entry is missing a few changes that are in this PR (Decision-to-Discard rewrite, payload-SHOULD relocation, Appendix B informative, RFC 7518/9864 refs, spec.yaml `jwksUri`, Trusted Server wording).

Also, the "Added informative aids" bullet right after this (same-string note, Appendix D label fix, Signing Direction Index, Appendix E, the diagrams) looks like #390 work to me. I could not attach a suggestion to it because those lines fall outside this PR's diff, but could you please move it to #390's changelog, or drop it from here?

```suggestion
## Version 07
* Required the `keyid` signature parameter and that it matches the
  `kid` of the verification key in the signer's JWK Set.
* The HTTP Message Signature algorithm is now derived from the JWK
  identified by `keyid`, per Section 3.3.7 of [RFC9421], instead of
  being restricted to the "HTTP Signature Algorithms" registry; the
  `alg` signature parameter is optional and checked for consistency.
* Replaced the unregistered `/.well-known/jwks.json` endpoint with a
  `jwksUri` field in the Discovery response, from which the location
  of the JWK Set for HTTP Message Signatures is discovered.  Added
  the matching `jwksUri` field to the OpenAPI specification.
* The `Date` header is no longer covered by signatures: freshness is
  anchored on the `created` signature parameter (see Section 7.2.4
  of [RFC9421]).
* Updated the signature examples: replaced the legacy `Digest` header
  with `Content-Digest` [RFC9530] and applied [RFC8792] line wrapping.
* Moved the JWKS-publication and payload-consistency requirements
  into the normative HTTP Message Signatures section, and marked
  Appendix B as informative.
* Rewrote the Decision to Discard text around the `http-sig`
  capability and the Trusted Server wording.
* Added normative references to [RFC7518] and [RFC9864].
```

IETF-OCM.md
```diff
@@ -827,12 +837,16 @@ components:
...
 
-A signed request SHOULD additionally cover the `date` component when a
-`Date` header is present.
+The `Date` header is deliberately not covered by the signature:
+intermediaries commonly rewrite it, which would make signatures
+fragile, and the `created` signature parameter already conveys the
+message's creation time (see Section 7.2.4 of [RFC9421]).
 
 The `content-digest` component binds the request body to the signature,
 protecting it against modification in transit.  Its value MUST use a
```
MahdiBaghbani, line 857: I think bodyless or empty-body requests could use one explicit line, so signers know to emit a Content-Digest of the empty body plus `Content-Length: 0` rather than dropping the header. Could we add a sentence after this paragraph to that effect?

IETF-OCM.md
```diff
@@ -827,12 +837,16 @@ components:
 * "content-digest"      - [RFC9530] digest of the body
 * "content-length"      - message size
 
-The Signature-Input parameters MUST include `created`.  Freshness and
-replay protection are anchored on `created` (see Verification
-Requirements).
+The Signature-Input parameters MUST include `created` and `keyid`.
```
MahdiBaghbani: Tiny terminology one: RFC 9421 calls these "signature parameters", not "Signature-Input parameters". I think matching the RFC wording avoids a reader looking for a distinct concept.

```suggestion
The signature parameters MUST include `created` and `keyid`.
```

IETF-OCM.md
```diff
@@ -827,12 +837,16 @@ components:
...
+Freshness and replay protection are anchored on `created` (see
+Verification Requirements).  The `keyid` value MUST be equal to the
+`kid` value of the corresponding key in the signer's JWK Set (see
+[Keys and Algorithms](#keys-and-algorithms)).
 
-A signed request SHOULD additionally cover the `date` component when a
-`Date` header is present.
+The `Date` header is deliberately not covered by the signature:
+intermediaries commonly rewrite it, which would make signatures
```
MahdiBaghbani: Two small notes here, neither blocking:

- The "intermediaries commonly rewrite it" gloss is fine, but it is not cited to any RFC section. The Section 7.2.4 cite right after is correct, so maybe just keep that and drop the uncited claim, or point it at 7.2.4 too.
- Only noting: freshness rests on `created` alone, so a captured async notification is replayable within the freshness window. If a nonce or `expires` is out of scope for now, I think a one-line note would stop implementers assuming replay is fully handled.

IETF-OCM.md
MahdiBaghbani, line 881: Only a note: `EdDSA` satisfies the MUST (asymmetric and registered) but not the SHOULD (fully-specified), and I understand real deployments use `alg: "EdDSA"`. The `Ed25519` RECOMMENDED covers the common case, so I am not asking for a change, just flagging it.

IETF-OCM.md
```diff
@@ -2008,6 +2074,10 @@ https://datatracker.ietf.org/doc/html/rfc9530)", February 2024.
 Representation of Contact Data](
 https://datatracker.ietf.org/doc/html/rfc9553), May 2024"
 
+[RFC9864] Jones, M., Steele, O., "[Fully-Specified Algorithms for
+JOSE and COSE](https://datatracker.ietf.org/doc/html/rfc9864)",
+October 2025.
```
MahdiBaghbani, line 2094: I think the RFC 9864 title is abbreviated. The published title is "Fully-Specified Algorithms for JSON Object Signing and Encryption (JOSE) and CBOR Object Signing and Encryption (COSE)".

```suggestion
[RFC9864] Jones, M., Steele, O., "[Fully-Specified Algorithms for
JSON Object Signing and Encryption (JOSE) and CBOR Object Signing
and Encryption (COSE)](https://datatracker.ietf.org/doc/html/rfc9864)",
October 2025.
```

IETF-OCM.md
```diff
@@ -2046,19 +2120,23 @@ out of scope for this specification: a mechanism similar to the
 
 # Appendix B: JWKS and HTTP Signature Examples
 
+This appendix is informative.
+
 ## JWKS Endpoint
```
MahdiBaghbani: The heading still says "JWKS Endpoint", but this is now a field rather than a fixed endpoint. I think "Published JWK Set" matches the new model better.

```suggestion
## Published JWK Set
```

IETF-OCM.md
```diff
@@ -2099,46 +2177,39 @@ Content-Digest: sha-256=:LkpHyFOVbBDPxc7YbHDOWNzAv88qWuVfLNf4TUf9Uo8=:
...
-"date": [date]
-"@signature-params": ("@method" "@target-uri" "content-digest"
-    "content-length" "date");
-    created=[timestamp];
-    keyid="sender.example.org#key1";
-    alg="ed25519";
-    tag="ocm"
+"@signature-params": ("@method" "@target-uri" "content-digest" \
+  "content-length");created=[timestamp];\
+  keyid="sender.example.org#key1";alg="ed25519";tag="ocm"
```
MahdiBaghbani, line 2206: A few small ones on the example, none blocking:

- The signature base mixes real header values with `[digest-value]` / `[body-length]` placeholders, so the example is half real. My preference would be placeholders throughout for the base and real values only in the headers.
- The Token example has no signature-base block of its own, and `@target-uri` uses the opaque `{tokenEndPoint}`, so a reader cannot resolve it. A small token signature-base block with a resolvable URI would round it out, but maybe later on another PR.
- Only noting: `created=1730815200` is 2024-11-05 while the old `Date` was in 2025. No live contradiction after the PR, but a consistent timestamp would read cleaner.

IETF-OCM.md
```diff
@@ -2099,46 +2177,39 @@ Content-Digest: sha-256=:LkpHyFOVbBDPxc7YbHDOWNzAv88qWuVfLNf4TUf9Uo8=:
...
-which is part of the signature base above:
+signature, and then add headers.  Note that the dictionary label
+(`sig1` below) is arbitrary; the signature is marked as belonging
+to OCM by its `tag="ocm"` parameter, which is part of the signature
+base above:
 
 <sourcecode type="http">
+NOTE: '\' line wrapping per RFC 8792
+
 Content-Digest: sha-256=:[digest-value]:
```
MahdiBaghbani: Only a nit: the Token `Content-Digest` line is around 70 chars and may trip an xml2rfc 69-column warning. Probably fine, just flagging.

IETF-OCM.md
```diff
@@ -2099,46 +2177,39 @@ Content-Digest: sha-256=:LkpHyFOVbBDPxc7YbHDOWNzAv88qWuVfLNf4TUf9Uo8=:
...
-Date: [date]
-Signature-Input: sig1=("@method" "@target-uri" "content-digest"
-    "content-length" "date");
-  created=[timestamp];
-  keyid="sender.example.org#key1";
-  alg="ed25519";
-  tag="ocm"
+Signature-Input: sig1=("@method" "@target-uri" "content-digest" \
+  "content-length");created=[timestamp];\
+  keyid="sender.example.org#key1";alg="ed25519";tag="ocm"
```
MahdiBaghbani, line 2222: The Token `Signature: sig1=:bM2s...hC=:` value is not valid base64 and not a 64-byte Ed25519 signature. I know this is pre-existing and the PR only rewraps it, but if it is easy, a real or clearly illustrative value would avoid someone copy-pasting it.

MahdiBaghbani (changes requested):
Thanks for this, I think the direction is really good.

Requiring `keyid`, deriving the algorithm from the JWK, and dropping `Date` from the covered components all make the signing model a lot easier to reason about. I also like moving the JWKS publication and payload-consistency requirements into the normative section and marking Appendix B as informative.

I have included suggested changes inline for each comment, so you can apply them quickly from the GitHub UI if you agree with them.

I left a few inline comments. Most of them are editorial, but three are worth sorting out before RFC last call:

- `jwksUri` has no https requirement (spec.yaml + I-D), and the I-D text does not say the JWK Set must be fetched over https.
- There is no `kid` uniqueness MUST within a JWK Set, so a verifier that hits two keys with the same `kid` has no defined behavior.
- The cross-registry `alg` mapping is only shown for Ed25519, so "MUST denote the same algorithm" is not really interoperable for RSA-PSS or ECDSA yet.

A couple of non-inline notes:

- The Version 07 changelog looks like it mixes in a few items that I think actually landed in #390 (the same-string note, the Appendix D label fix, the Signing Direction Index, Appendix E, the diagrams). Could you please double-check those belong here?
- Removing the fixed `/.well-known/jwks.json` path is a real wire change for peers that hardcoded it, and it adds a mandatory Discovery round-trip. I think a short migration note would help implementers.
- The diagrams under `diagrams/` (02-discovery-receiver-publish, 07-signing-directions) still hardcode `/.well-known/jwks.json`. I added those diagrams myself, so if you like I can fix the diagram paths in this PR too, to keep them consistent with the new `jwksUri` model.

A few more small ones I could not attach inline because the lines fall outside this PR's diff:

- spec.yaml line 32: the 200 description still says "public key" (singular); now it is a JWK Set via `jwksUri`.
- IETF-OCM.md line 1348 and line 1999: "HTTP Signature" should read "HTTP Message Signature(s)" to match the section title and RFC 9421.

The examples check out as far as I can tell: the Content-Digest values and content-lengths match the example bodies, and the RFC 8792 folding looks mechanically correct.

Some of my suggestions are a bit more normative than editorial, so if I misunderstood the intended model, feel free to push back and we can narrow it down.

IETF-OCM.md
MahdiBaghbani: Or not! since almost nobody uses this today?
I mean I haven't had the time to check any of them yet

IETF-OCM.md
MahdiBaghbani: Here is a real signature that verifies against `x: xofVboLdrD4KBvul7AUxUccIGeB-5CBmlp37UHzgEVw` (a fresh key for `receiver.example.org#key1`), over the example base with `{tokenEndPoint}` resolved to `https://cloud.example.org/ocm/token`.

```suggestion
Signature: sig1=:pnDNnZAvDsTFTTt56dC5TpL99vFvGnC3hTGsi4OO2TownX3AiIDnhUBZOWr2UpPctT4Yb\
  /XHr9dhryo2Lyn1Bg==:
```

The `POST {tokenEndPoint}` line probably wants the concrete URL too, so the base is reproducible from the headers.

IETF-OCM.md
glpatcern: Yeah to be honest I'm not sure we want to go ahead with this right away, at least I'd wait what the JOSE WG would tell us about their (lack of) well-known URL.

IETF-OCM.md
glpatcern: > could you please move it to #390's changelog, or drop it from here?

Mahdi, prior to merging your work I moved the related changelog entries to version 07 because 06 was already out, so it's all fine. We will have soon a 07 I guess...

MahdiBaghbani: Ah right!

IETF-OCM.md
mickenordin: I am not sure we should do that, RFC7517 describes scenarios where they MAY be the same.

https://www.rfc-editor.org/info/rfc7517/#section-4.5

IETF-OCM.md
mickenordin: I don't think we should have a table tbh. The algorithms mentioned in RFC9421 are only the intial content of the registry and may thus change in the future, making this list stale in our RFC: https://www.rfc-editor.org/info/rfc9421/#iana-hsa-contents

IETF-OCM.md
mickenordin: Like I said above, I think we should allow multiple keys as spelled out in RFC7517

IETF-OCM.md
mickenordin: I don't know that we do any signed request with an empty body, or even signed get-requests?

IETF-OCM.md
MahdiBaghbani: You are right, RFC 7517 only makes distinct kid a SHOULD.

The part I think needs more info is that OCM picks the key by keyid == kid, so a duplicate kid leaves the verifier with no deterministic pick. If same kid stays allowed, maybe we just reject an ambiguous kid, or pick by alg and reject if more than one matches.

IETF-OCM.md
MahdiBaghbani: Fair point, all signed OCM requests today are POSTs with a body, I'm over thinking this one 😃

IETF-OCM.md
mickenordin: I think the fact that there is no .well-known address registered for this defeats the purpose of using one. The reason to use it was that people would just be able to reuse their current deployment off the shelf. But since there is no such standard way, it is actually better not to invent one. OAuth has jwks_uri in their discovery, and I think mimicing this, is what will allow people to reuse their current deployments.

So we should do the same thing that others do, rather than pushing fo a new way of doing it, in this spec, or elsewhere, when there is already prior art.

IETF-OCM.md
mickenordin: IIRC 7.2.4 does not say anything about rewriteing the date, this wording came from the reviewer and not from the RFC so I could not point to it for that.

IETF-OCM.md
mickenordin: The column limit is 72 chars I think.

IETF-OCM.md
mickenordin: And as you pointed out, Nextcloud is the only implementation yet, so when I fix that it will be compliant. We can still serve the jwks.json from the same place, we just need to augment the discovery.

glpatcern: Eventually the discussion in the JOSE mailing list made it very clear that this is the way to go, let's go then!

mickenordin:
I think I addressed your feedback now @MahdiBaghbani

glpatcern (dismissed):
This looks good now, Mahdi you have the last word following your detailed review

MahdiBaghbani:
> This looks good now, Mahdi you have the last word following your detailed review

I'll review this soon

mickenordin:
Maybe we shouldn't mandate HTTPS for fetching the key material @MahdiBaghbani ?

It breaks real deployments in Nextcloud, as well as a lot of testing setups: https://github.com/nextcloud/server/pull/62543?email_source=notifications&email_token=AAJY6JQRXSIFSNYUSG6UXFL5GZQ2HA5CNFSNUABFM5UWIORPF5TWS5BNNB2WEL2JONZXKZKDN5WW2ZLOOQXTKMBYGUYTSOJUGI4KM4TFMFZW63VGMF2XI2DPOKSWK5TFNZ2KYZTPN52GK4S7MNWGSY3L#issuecomment-5085199428

I would be content with RECOMMENDING https.

MahdiBaghbani:
Fair point, and thanks for the link. I don't want to break Nextcloud or people's test setups over this, so let's find something that works for both.

The part that was making me to think is the threat model. Discovery already comes over https, so we get the `jwksUri` string with integrity. But if that string is an `http://` URL, someone on the path who can't touch your https discovery can still MITM the key fetch and swap the JWK. Then they sign whatever they like and we verify it happily. So for jwksUri specifically the https is doing real work, it's not cosmetic.

Funny thing is the prior art you pointed at agrees. RFC 8414 is the one that defines `jwks_uri` for OAuth discovery, and it says:

> This URL MUST use the "https" scheme.

OIDC Discovery is the same, and its section 7.2 describes exactly this attack, an RP "tricked into using an attacker's keys and endpoints". So mimicking OAuth here actually means keeping the MUST 😃

But I suspect the thing that's really biting you isn't the prose, it's `spec.yaml`:

```yaml
pattern: "^https://"
```

That's a hard schema failure, so an `http://` test setup can't even validate. That one I'd happily drop.

What I'd suggest: keep the MUST in the I-D, and give it the same testing escape hatch discovery already has in Step 4, "Implementations MAY fallback to HTTP instead of HTTPS in testing setups". Then remove the `pattern` from spec.yaml and let the prose carry it. Production stays safe, your test setups and the Nextcloud PR keep working.

Would that work for you? If you still prefer RECOMMENDED I'm not going to block it, I'd just want a line in Security Considerations saying what we give up.

One more while we're here, `tokenEndPoint` has the opposite gap: the I-D says those requests MUST be over HTTPS (line 2002) but spec.yaml has no pattern at all. However we land jwksUri, those two should probably match.

IETF-OCM.md
```diff
@@ -827,12 +837,17 @@ components:
...
 * "content-length"      - message size
 
-The Signature-Input parameters MUST include `created`.  Freshness and
-replay protection are anchored on `created` (see Verification
-Requirements).
+The signature parameters MUST include `created` and `keyid`.
+Freshness and replay protection are anchored on `created` (see
+Verification Requirements).  The `keyid` value MUST be equal to the
+`kid` value of the corresponding key in the signer's JWK Set (see
+[Keys and Algorithms](#keys-and-algorithms)).
```
MahdiBaghbani, line 848: Small one, but worth fixing since it's a security claim.

`created` gives us freshness, but not replay protection. Inside the acceptance window an attacker can replay a captured request as many times as they like and every signature still verifies. RFC 9421 splits these two pretty explicitly in 7.2.2:

> the signature can use the nonce signature parameter to provide a per-message unique value to allow the verifier to detect replay of the signature itself if a nonce value is repeated. Furthermore, the signer can provide a timestamp for when the signature was created and a time at which the signer considers the signature to be expired, limiting the utility of a captured signature value.

So `created` limits the utility, `nonce` is what actually detects it. We have no `nonce` and no replay cache in the draft, so as written we're promising something we don't have.

I'm not asking to add nonce in this PR, that's a bigger conversation. Just that the text shouldn't claim it:

```suggestion
The signature parameters MUST include `created` and `keyid`.
Freshness is anchored on `created` (see Verification Requirements);
`created` bounds how long a captured signature stays acceptable, but
does not by itself detect replay within that window (see Section
7.2.2 of [RFC9421]).  The `keyid` value MUST be equal to the
`kid` value of the corresponding key in the signer's JWK Set (see
[Keys and Algorithms](#keys-and-algorithms)).
```

IETF-OCM.md
```diff
@@ -2148,24 +2229,30 @@ illustrates the procedure to verify an incoming signed request:
...
 1. Extract the provider domain from the `sender` field in the
    request body
-2. Fetch the public key from
-   `https://<provider-domain>/.well-known/jwks.json`
-3. Locate the unique signature carrying the `tag="ocm"` parameter in
+2. Fetch the Discovery response from
+   `https://<provider-domain>/.well-known/ocm` and read its
+   `jwksUri` field.  If the Sending Server advertises the `http-sig`
+   capability but no `jwksUri` is present, the receiver MUST reject
+   the request as non-conformant
```
MahdiBaghbani: This one's small but I think it'll bite an implementer.

The appendix is marked informative up at line 2132, but this step is a MUST, and it's a stronger MUST than the normative text. Decision to Discard says:

> The Receiving Server MAY discard the notification if any of the following hold true:
> * the Sending Server advertises the `http-sig` capability but its Discovery response carries no `jwksUri`

So normative says MAY discard, informative says MUST reject. Someone reading only the normative section can accept the request, someone reading the appendix has to reject it, and both of them are following the spec. And a MUST inside a section declared informative has no force anyway, so I'd expect the RFC Editor to flag it.

Easiest fix is to let the appendix describe the rule instead of inventing a stronger one:

```suggestion
   `jwksUri` field.  If the Sending Server advertises the `http-sig`
   capability but no `jwksUri` is present, the receiver can discard
   the notification, as described in [Decision to
   Discard](#decision-to-discard)
```

If you do want MUST reject, that's fine by me, but then it should move up into Decision to Discard, and that's a real behaviour change worth its own discussion.

MahdiBaghbani (changes requested):
Re-read the whole thing against the current head and it's in much better shape, thanks for working through all of that. Two small ones left below, and I answered the https question up in the thread.

Marking changes requested just to keep them on the radar, both are quick fixes and I've left suggestions you can click.

mickenordin:
I think this is good, accept http in testing setups.

I'll push another commit, probably tonight to fix it. If you approve of this, you can approve the pr, and I can merge after I have pushed the commit.


Den 27 juli 2026 11:21:27 CEST, Mahdi Baghbani ***@***.***> skrev:

MahdiBaghbani (approved):
Approving, this is good to go. Thanks for turning all of that around so fast.

PR #393, opened 2026-07-28
Add Apache 2.0 LICENSE

glpatcern:
This PR brings in License terms for the entire repository. Even though the IETF drafts "belong" to the IETF, the other material appears to require a license.

The text was copied verbatim from https://github.com/cs3org/reva/blob/master/LICENSE - we still have to validate this is correct.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,175 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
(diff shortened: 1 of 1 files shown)


MahdiBaghbani:
Specifications need license? Didn't know that!

Doesn't it technically belong to IETF now?

mickenordin:
What is the purpose of this? We already have the copyright sorted out, including a specific grant for copying the spec?

glpatcern:
It was asked by a CERN IP scrutiny as part of preparing for the EOSC4ALL project. I'm not convinced we need it, I've put it forward for discussion/evaluation.

mickenordin:
> It was asked by a CERN IP scrutiny as part of preparing for the EOSC4ALL project. I'm not convinced we need it, I've put it forward for discussion/evaluation.

What was the concern they put forward? CERN of course has a right to implement the specification by the standard IETF provisions and we have even granted further grants for software examples. So my take is that this will not improve the ability of anyone to implement or use the standard in anyway, but rather make it harder, to live up to more requirements. So I think we shouldn't do it, unless there are some clear benefits to the community.

If you want to proceed further with this, first read up on the IETF IP rules: https://www.rfc-editor.org/rfc/rfc5378.html and if you are not content with the status que, contact the IETF trust and ask them about this.

MahdiBaghbani:
My read is that we're not blocked. RFC 5378 section 3.6 says contributors, or their
employers, keep the copyright in Internet-Drafts, and the Trust only takes ownership
once something is published as an RFC. Section 5.10 then spells out that authors keep
the right to exploit their own contributions "in any way they wish". So the drafts are
still ours to license.

@mickenordin I think you're right that the drafts don't need it though. `IETF-OCM.xml`
carries `ipr="trust200902"`, so all three drafts are under BCP 78 and the TLP already,
code components pulled out of them are automatically Revised BSD (TLP 4.c), and on top
of that we added our own grant in Copying conditions that even allows modification. I
couldn't find anything missing there.

What I didn't expect is everything else in the repo: `spec.yaml`,
`schemas/ocm-discovery.json`, `diagrams/`, `docs.html`, `logo/`. None of them carry a
license or a copyright line as far as I can see.

The Copying conditions grant lives inside the draft and covers the draft. The TLP
covers IETF Documents. `spec.yaml` isn't either one, it's just a file in a GitHub repo,
so it falls back to plain default copyright, which means all rights reserved. And
`spec.yaml` is the file people actually feed to a generator to build a client. Same for
the schema.

So my guess is that's what CERN found, and if so I think it's a fair catch rather than
paperwork.

On the PR itself, could you please add a copyright line? Right now it's Apache 2.0
sections 1 to 9 with the APPENDIX cut off, so there's no "Copyright <year> <holder>"
anywhere, and Apache defines "Licensor" as the copyright owner granting the license. As
it stands I think the file grants rights on behalf of nobody, which would be a bit
funny to send back to an IP review.

My preference would be to scope it rather than drop it: have LICENSE say the
Internet-Drafts are governed by BCP 78 and the TLP, put the spec files and diagrams
under Apache 2.0, leave the logo out of it since Apache section 6 grants no trademark
rights anyway and add that copyright line.

One thing in favor of Apache specifically that I haven't seen come up: section 3 is an
express patent grant. Our Copying conditions grant doesn't have one, and neither does
the Revised BSD the TLP gives for code components. BCP 79 gets us disclosure, which
isn't quite the same thing. If someone's lawyer asks about patents before they
implement, I think that's the part that answers them. Without that, we could just as
well use Revised BSD and match the TLP.

I'm not a lawyer though 😄

lisad:
I am also not a lawyer, but I should think that applying a license to the
Github repo is going to be easier than trying to add it to the spec, and
that may help an org feel safer,  especially with examples or test code
that are alongside the spec.

Lisa

KrausMatthias:
Tangential, but it might be an option to use https://reuse.software/ to license specific parts of the repo

MahdiBaghbani:
@KrausMatthias That's right!

I have already done that in the https://github.com/MahdiBaghbani/opencloudmesh-go to learn and see how it goes, I would do it for this repo as well

mickenordin:
I would second @KrausMatthias proposal to add resuse comments to the specific files we want to licence under apache 2.0 and leave the I-D alone

mickenordin (changes requested):
Adding reuse comments to the specific files we want to licensens is better I think, since this implies a separate licence for the I-D drafts which we have already handled in the drafts themselvs

glpatcern:
I have clarified this with the relevant office at CERN, and the outcome is that we do not need to license the repository. I made it clear that the sources for the IETF datatracker are already the IETF IP (with us keeping authorship), and for the other assets (the diagrams and the logo) it's not really that important: we might want to "protect" the logo, but we could not even find the original author, so it's a bit pointless. Therefore I'm closing this PR.

PR #395, opened 2026-07-30
OCM-IP: resolve JWKS via Discovery jwksUri, not a fixed path

MahdiBaghbani:
The OCM-IP draft locates the OCM Server's signing keys at a fixed `/.well-known/jwks.json`, but the base OCM spec replaced that unregistered endpoint with the `jwksUri` field advertised in the Discovery response (see the Version 07 changelog).

Continuation of #391


mickenordin (approved):
Great

PR #397, opened 2026-08-09
Add MLS virtual client requirement for multi device support

mickenordin:
As evidenced from the WG meeting at IETF126, the OCM-MLS ID was too hand-wavy about exactly how a user could have multiple devices/native clients. This patch introduces a dependency on MLS Virtual Clients to bridge this gap: https://datatracker.ietf.org/doc/html/draft-ietf-mls-virtual-clients-01

I also fixed some broken formatting I found while reading/writing.


glpatcern (approved):
Thanks Micke, makes very much sense given the discussions at IETF.

A separate concern is about the maturity of the _virtual clients_ capability within MLS: hopefully we will be able to assess it and the whole approach once we get into implementing the protocol in the coming months.

PR #399, opened 2026-08-12
Added the spec and IANA registry entries for `embedded`

glpatcern:
With this PR I add a document that could serve as blueprint of how to extend the OCM IANA registries when it comes to new protocols and resource types. This document is independent from the I-D Drafts and may be moved elsewhere, but for now I think it makes sense to version-control it here.

The specific case is for an `embedded` protocol and an `ro-crate` resource type. This follows the discussion at https://github.com/cs3org/OCM-API/discussions/180 and it has been updated to the latest OCM spec version for Discovery. The underlying concept is that an `embedded` OCM Share may carry any resource that can be represented in JSON, with `ro-crate` being a specific case.

The text has been written with some AI assistance for the general structure.

### DIFF
--- a/OCM-Embedded.md
+++ b/OCM-Embedded.md
@@ -1,5 +1,11 @@
 # Embedded JSON Resources in Open Cloud Mesh
 
+**Giuseppe Lo Presti**<sup>1</sup>, **Rasmus Oscar Welander**<sup>1</sup>
+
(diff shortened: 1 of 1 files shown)


glpatcern:
@KrausMatthias would you have time to take a look and sketch a similar document for your GIS-related project?

mickenordin (approved):

PR #400, opened 2026-08-17
Notifications rehaul

glpatcern:
Main areas:
- Notification payloads are fully specified in case of `file` resources or shareType (`user`, etc.) recipients
- Request to share is now a notification
- Reshare operations are better explained, including the fact that access must always happen to the owner's server, not via third parties. At the moment, this implies that even if Bob received a share with `share` permission, Bob cannot send the share to Charlie right away and MUST send Alice a `REQUEST_RESHARE` notification, to ask Alice to share the same resource to Charlie.

Still missing:
- The format of a notification response when successful is left unspecified
- A `FEDERATION_REMOVED` notification is not specified, do we need it?
- The payloads of each notification type are specified in the I-Ds but not in `spec.yaml`. To be seen if we want to have satellite OpenAPI specs for them.

I acknowledge this is quite a large single commit. If needed I can extract the request to share part, but as it came through the editing it was not obvious to take it apart.

### DIFF
--- a/IETF-OCM.md
+++ b/IETF-OCM.md
@@ -1342,20 +1342,27 @@ request:
 * REQUIRED notificationType (string) - it MUST be one of the
   registered values listed in the "OCM Notification Types"
   registry (see [IANA Considerations](#iana-considerations)).
-* REQUIRED senderDomain (string) - ...
-* REQUIRED resourceType (string) - ...
+* REQUIRED senderDomain (string) - the FQDN of the sender.  A Receiving
+  Server SHOULD use this information to verify the HTTP message
+  signature on the request, and it SHOULD also use it to validate the
+  underlying share or resource the notification is about.
+* OPTIONAL resourceType (string) - the type of the resource this
+  notification is about (e.g., `file`).
+* OPTIONAL shareType (string) - the type of recipient this notification
+  is about (e.g., `user`).
 * OPTIONAL providerId (string) - the identifier assigned by the
   Sending Server to the underlying share, if applicable. This
   field is _deprecated_ and SHOULD NOT be used.
 * OPTIONAL notification (object) - optional additional parameters,
   depending on the notification and the resource type.
 
-Multiple Notification types are defined for different purposes, and
-additional types MAY be defined by implementers and registered in the
-"OCM Notification Types" IANA Registry.  For each notification type,
-a specific format is defined for the optional notification object.
-In the sections below the base Notification types are detailed along
-with their specific payload.
+A Notification payload MUST include either the `resourceType` or the
+`shareType`, or both when relevant.  Multiple Notification types are
+defined for different purposes, and additional types MAY be defined
+by implementers and registered in the "OCM Notification Types" IANA
+Registry.  For each notification type, a specific format MAY be defined
+for the optional notification object.  In the sections below, the base
+Notification types are detailed along with their specific payload.
 
 ## Share Acceptance and Updating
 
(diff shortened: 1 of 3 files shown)


glpatcern:
I shall add specific review comments to the changes proposed here, and I'm happy to receive specific feedback (I anticipate I don't like a couple of them). But meanwhile, as a preview of the proposed changes, I've updated the presentation I gave at the IETF WG meeting: please find it at https://cernbox.cern.ch/s/zEMiLaFRLJFrS4y

cc @ArtificialOwl for Nextcloud, @kobergj for ownCloud, @butonic for OpenCloud.

And a gentle reminder: during the OCM calls, as well as during the IETF meeting, we iterated on the fact that the new format _has to_ be different, thus incompatible, from the current, **undocumented** formats adopted by Nextcloud and oCIS/OpenCloud, which are NOT interoperable. In other words, I call on all implementations to look into the necessary adjustments, starting from a simple check, in pseudo-code:
```
if notification['senderDomain'] is defined {
    parse new-format notification
}
else {
    parse old-format, implementation-specific notification
}
```

IETF-OCM.md
```diff
@@ -1380,35 +1339,155 @@ Receiving Server MAY make a HTTP POST request
...
+
+## Share Acceptance and Updating
+
+A notification MAY be sent by a recipient to let the provider know that
+the recipient accepted or declined a share, in response to a Share
+Creation Notification.  Similarly, it MAY be sent by a provider to let
+the recipient know that the provider updated or removed a given share,
+such that the recipient MAY clean it up from its database.  In all such
+cases, the `resourceType` is expected to match the `resourceType` of
+the underlying share (e.g., `file`), and the `notificationType` MUST be
```
mickenordin: we should add a comment here that explicitly states that all examples for `file` below are also valid for `folder` so we don't need to repeat them for folder too.

mickenordin (changes requested):
Very good! Only one single nitpick.

Since we define both file and folder in this document, we should fully specify the behaviour for folder too, although I think a simple sentence sayint that it is the same as for file suffices.

IETF-OCM.md
glpatcern: Fully agreed, will push an amended version of the text

MahdiBaghbani (dismissed):
For me it's quite a good start!

IETF-OCM.md
```diff
@@ -1380,35 +1339,155 @@ Receiving Server MAY make a HTTP POST request
...
-  - 'SHARE_DECLINED'
-  Registered values are listed in the "OCM Notification Types"
+* REQUIRED notificationType (string) - it MUST be one of the
+  registered values listed in the "OCM Notification Types"
   registry (see [IANA Considerations](#iana-considerations)).
-* REQUIRED providerId (string) - copied from the Share Creation
-  Notification for the Share this notification is about
-* OPTIONAL resourceType (string) - copied from the Share Creation
-  Notification for the Share this notification is about
+* REQUIRED senderDomain (string) - ...
```
mickenordin: Here we should say that the recipient MAY use this value to verify the HTTP message signature on the request, but SHOULD also take care to tie the domain to the share before altering the state. I.e. double check that the domain has not been spoofed, but it can be done after the notification has been received, when it is being processed for the content. E.g Nextcloud does not know which share a notification belongs too in the api controller, so now we have to resolve the share directly when it is comming in, rather than later when we know which share it is.

mickenordin (approved):

PR #402, opened 2026-08-28
Move encryption to the main I-D

### LINKED ISSUE #401: Need to move `encryption` from OCM-MLS I-D to main I-D
While reviewing the specs for Notifications in #400, I saw that we describe an optional `encryption` block in a Share Creation Notification payload as part of OCM-MLS, but there's no reference to it in `spec.yaml`.

I think we should move the block to the main I-D as well as describe it in `spec.yaml`. From discussions at IETF, it was evident that options for encryption may be useful in any OCM share, not just to support the federation.

glpatcern:
Initial attempt at moving the `encryption` object spec without altering the semantic for MLS

### DIFF
--- a/IETF-OCM-MLS.md
+++ b/IETF-OCM-MLS.md
@@ -1587,19 +1587,22 @@ Each notification MAY include the optional `encryption` field:
   },
   "encryption": {
     "resourceId": "3a02538b-aa54-42f2-8853-a38996e211b1"
-    "scheme": "ocm-mls-1",
-    "cipher": "AES-256-GCM",
+    "scheme": "ocm-mls",
+    "ocm-mls": {
+      "cipher": "AES-256-GCM",
+    }
   }
 }
 ~~~
 
 The `encryption` field is OPTIONAL as per [OCM] Share Creation
 Notification specification.  If absent, the resource is unencrypted.
-If present, it MUST carry the encryption-related parameters according
-to the [OCM] specification, where:
+If present, and the symmetric encryption available via OCM-MLS is to
+be used, it MUST carry the encryption-related parameters according to
+the [OCM] specification, where:
 - `resourceId` is the stable resource identifier described in
 {{resource-id}}
-- `scheme` is to be set to `"ocm-mls-1"`
+- `scheme` is to be set to `"ocm-mls"`
 - `cipher` names the content AEAD that the resource is encrypted with
 ({{file-key-wrapping}}), one of the AEAD algorithms defined for HPKE
 ([RFC9180] Section 7.3): `"AES-128-GCM"`, `"AES-256-GCM"`, or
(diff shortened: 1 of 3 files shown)


IETF-OCM-MLS.md
```diff
@@ -1589,20 +1586,21 @@ Each notification MAY include the optional `encryption` field:
...
-The `encryption` field is OPTIONAL.  If absent, the resource is
-unencrypted and the share follows the standard OCM flow without
-modification.  If present, it carries all encryption-related parameters:
-`scheme` identifies the encryption scheme, for which this document
-defines `"ocm-mls-1"`; `resourceId` is the stable resource identifier
-described in {{resource-id}}; and `cipher` (REQUIRED when `encryption`
-is present) names the content AEAD that the resource is encrypted with
+The `encryption` field is OPTIONAL as per [OCM] Share Creation
+Notification specification.  If absent, the resource is unencrypted.
+If present, it MUST carry the encryption-related parameters according
```
mickenordin:
```suggestion
If present, and you want to use the symmetric encryption available via OCM-MLS, it MUST carry the encryption-related parameters according
```

mickenordin: I mean that it should be possible to send a federated share using public key encryption too.

IETF-OCM.md
```diff
@@ -1092,6 +1092,17 @@ described in [OCM-IP].
...
   the resource represents a cached copy of a dataset that was made
   available for an efficient data transfer to the destination server.
+* OPTIONAL encryption (object)
+  Optional JSON object with encryption information for the share.  If
+  omitted, it is assumed that the resource is not encrypted.  For
+  encrypted resources, the actual key material is to be exchanged out
+  of band, and this object MUST include:
+  - REQUIRED resourceId (string) - a unique identifier of the
+    underlying resource.
+  - REQUIRED scheme (string) - an identifier of the encryption scheme
```
mickenordin, line 1102: for ocm-pgp-1 I think we need either fingerprints or recipients (in the pgp sense, not the ocm sense) so the recipient knows which key to use for decryption

spec.yaml
```diff
@@ -603,6 +603,31 @@ components:
...
+                The unique identifier of the underlying resource. This MUST be the same
+                for all shares of the same resource.
+            scheme:
+              type: string
+              description: >
+                An identifier of the encryption scheme used to encrypt the resource.
+            cipher:
+              type: string
+              description: >
+                The encryption algorithm used to encrypt the resource.
```
mickenordin: Same here, I think we need to somehow point to the keys that can decrypt, I think a `fingerprints` array is appropriate for ocm-pgp-1 scheme.

mickenordin (changes requested):

mickenordin (approved):

mickenordin:
Looks good!

MahdiBaghbani (approved):
Great 🥳

PR #403, opened 2026-08-30
Improved phrasing for jwksUri, fixes #394

### LINKED ISSUE #394: Clarify: must `jwksUri` be same-authority as the discovery endpoint, or is cross-host allowed?
Hi all, I'm working on an OCM implementation and ran into something the spec doesn't seem to answer directly, so I wanted to get other people's read on it before I lock anything in.

When a server advertises `jwksUri` in its discovery response, does the `jwksUri` host have to match the discovery/ocm endpoint host, or is a different host allowed (say a CDN or a dedicated keys host)?

The spec is clear about three things: the path is not fixed, it MUST use https (HTTP only in testing), and it MUST be advertised when `http-sig` is exposed. What I can't find is any line saying the `jwksUri` host has to equal the discovery host. The example is same-host with a custom path (`https://cloud.example.org/ocm/jwks`), but that's an example, not a constraint.

Refs at `6a05861`:
- spec.yaml: https://github.com/cs3org/OCM-API/blob/6a0586183cbef10ecae9dedc42561806447eb2f5/spec.yaml#L534-L545
- discovery field: https://github.com/cs3org/OCM-API/blob/6a0586183cbef10ecae9dedc42561806447eb2f5/IETF-OCM.md#L774-L782
- verifier discard reasons: https://github.com/cs3org/OCM-API/blob/6a0586183cbef10ecae9dedc42561806447eb2f5/IETF-OCM.md#L1297-L1299

That last one ties the JWK Set to "the FQDN part of the `sender` field ... via the `jwksUri` field of its Discovery response". To me that reads like the keys belong to the same server identified by the sender FQDN, but it doesn't actually add a "host must match" rule, it just says the keys are obtained via whatever URL the discovery advertises.

I can argue both ways and I'm honestly not sure which is intended.

For allowing cross-host: the spec doesn't forbid it, and moving `jwksUri` into a discovery field was supposedly about operator flexibility. A CDN or a separate keys host is a real pattern, OIDP IdPs do it with `jwks_uri` all the time. And for self-advertisement it's the operator's own URL, not peer-controlled input, so the SSRF argument against it is kind of weak.

For same-authority only: the FQDN -> discovery -> `jwksUri` chain reads more naturally if the keys live on the same host. And if a receiver enforces same-authority on peers (some do), a server that advertises a cross-host `jwksUri` just becomes interoperably broken, peers would reject the URL it advertised. Same-host also keeps the fetch target on the already-vetted discovery origin, smaller surface than "any https URL a discovery doc points at".

So, a few things I'd like to hear from you:

1. Is same-authority for `jwksUri` an intended constraint, or just a side effect of the example?
2. How do existing implementations (Reva, Nextcloud, ownCloud/oCIS) treat a peer `jwksUri` whose host differs from the discovery host? Accept, reject, or not implemented yet?
3. Worth adding one normative line either way, a MUST for same-authority or an explicit "host is not constrained" note?

glpatcern:


MahdiBaghbani (approved):
Thanks!

Fixes: #394

mickenordin (approved):

PR #411, opened 2026-08-31
Fix OCM IP example and JSON commas

glpatcern:
Both issues were found when generating the XML for the datatracker, see #405


MahdiBaghbani (approved):
looks good to me!

